PR #8, opened 2021-04-27
FDIP: prefetch multiple cache lines if bbl spans line boundaries

petervbraun:
Prefetch multiple cache lines if basic block spans across cache line boundaries


hlitz (commented):
can you merge your multiple commits into a single commit (in a new branch) so that I can pull a single clean commit?

PR #138, opened 2025-01-09
Fix clang-format to follow Google's convention

Kiminje:
ReflowComments: true -> wrap long comments to fit in ColumnLimit(unfortunately, cannot merge shorter multiple comments)
PointerAlignment: Right -> pointer alignment
SpaceAfterCStyleCast: false -> formerly Google allow space for C-style case for readability, but not now. 
AllowShortFunctionsOnASingleLine: Inline -> now single line function in class can be inlined. Otherwise, short functions outside of classes are changed their line


hlitz (approved):
LGTM

PR #209, opened 2025-03-18
Add features: detect misprediction by reason

Kiminje:
To use perfect tag mode, add '--use_perfect_tag 1' to your config file.

Background
TAGE has 2 tables: one for short history lengths and one for longer history lengths. Group of history lengths share TAGE table. For example, the TAGE table for short history lengths has 10 banks, and each bank has 1024 entries. Notice that the banked structure allows parallel access to history lengths. Each history length accesses the table with the given hashed index(index's range is all entries and banks) and tag. Replacement policy:
Based on matched history length, the TAGE predictor traverses the table with longer history lengths 'i' and given index 'GI[i].' Evicted entry is selected when the entry's u-bit is 0, and ctr is weak(abs(2*ctr + 1) <= 3).

Motivation
To reveal mispredictions from TAGE by reason, add collecting metrics and a live_perfect_tag_map.

Implementation
TAGE entry has an additional member, perfect_tag, to detect tag aliases. Now GTAG[i] has perfect_tag initially, and the masked tag will be stored at update(if --use_perfect_tag is disabled). All the mispredictions are collected only for on-path ops.

Entry miss: since TAGE is accessed by a given index, it's hard to detect capacity miss. So, this PR collects only cold and conflict miss. PR, which evaluates TAGE's capacity pressure needs.

ENTRY EXIST AND HIT
If an entry exists, the correct entry is distinguished by 'CORRECT_TAGALIAS' and 'CORRECT'. 'CORRECT_TAGALIAS' means the entry is correctly predicted, but the tag is aliased. Otherwise, it's purely correct.

ENTRY EXIST BUT MISPREDICT
For misprediction, three cases exist: MISPR_COLD_TAG_ALIAS, MISPR_TAGALIAS, and MISPR_MISPRED.
1. This PC should be cold miss, but entry hits due to tag aliasing, and it leads to misprediction.
2. The entry's tag is aliased and mispredicted.
3. None of the above cases are purely mispredicted by TAGE.

ENTRY MISS
if not cold miss, all of the cases are conflict miss

LIVE_PERFECT_TAG_MAP
This is to store the perfect tag of current TAGE entries. live_perfect_tag_map stores map<index, counter> and are accessed by perfect_tag. When TAGE allocates a new entry, live_perfect_tag_map also evicts the corresponding index's entry and allocates a new entry with the given perfect_tag. Counter in live_perfect_tag_map is updated when TAGE updates, but the entries that match with GTAG[idx] are updated while gtable is updated by the given index. This is to evaluate how effective it will be to use live_perfect_tag_map to remove mispredicted tag_aliased entries. Since allowing tag_aliasing for the current design is a cost and performance-effective methodology, we need more observations to improve the TAGE predictor. live_perfect_tag_map collects metrics for the cases in which an entry hit but was mispredicted or an entry missed. Using the longest index's counter in matched perfect_tag, it provides prediction and records the hit rate.

No perf Drift


hlitz (changes requested):
Hi,
I tried to review but its hard to determine whether its correct for me. I would suggest a simpler approach: Have a knob to enable perfect tag, infinite size. Then run the apps once with enabled perfect tag and once without and then one can compare the results. Or do you think your current method of doing it at the same time is more powerful? 

In my proposal I would expect to see an if(PERFECT_TAG) whenever you read or write the tag but the rest of the code unmodified.

PR #213, opened 2025-03-20
Enable scarab test simulation run after build

hlitz:


5surim (changes requested):
Hi @hlitz ,

Thank you so much for the CI to test simulation run.
Will the performance_log.csv be available in scarab_ll repository?

Surim

PR #228, opened 2025-03-31
Add support for Google v.2 memtraces. 

hlitz:
Google memtraces are filtered and hence do not contain actual x86 opcodes that could be parsed with Intel XED.
To address this issue, this patch parses the available dynamorio information and directly generates instructions of type
ctype_pin_inst in trace_memtrace_reader_memtrace.cc.


5surim (changes requested):
Hi @hlitz ,

thank you for the patch.
It seems the order of the new stat should be fixed.
Others look good to me.

Surim

PR #316, opened 2025-05-14
[Feature] Perceptron-Based Confidence

naomirehman1008:
#### Summary
This PR introduces a confidence mechanism using a table of perceptrons as a predictor. The table is indexed with PC^HIST. It uses saturating counters for weights.

#### Parameters
- **CONF_PERCEPTRON_CONF**: Used to enable perceptron-based confidence
- **CONF_PERCEPTRON_THRESHOLD**: Threshold for predicting off-path. Used to move the decision boundary to favor on or off-path predicitons
- **PATH_CONF_PERCEPTRON_LR**: Learning rate.
- **PATH_CONF_N_PERCEPTRONS**: Number of rows in the table of perceptrons.
- **CONF_PERCEPTRON_THETA**: Threshold for training. If the absolute value of the prediction sum is lower than theta, the perceptron will train reguardless of whether the prediction was correct or incorrect.
- **CONF_PERCEPTRON_WEIGHT_WIDTH**: Bit width of weights.

#### Training
The table is trained when a cf op retires or when a cf op is invalidated on an FTQ flush. It is trained with the same data as when the op was used to produce a prediction. The prediction, prediction value, and features are kept in the op struct for training. 

#### Features
There are 22 features. Each feature has a flag that can be used to disable/enable it's use in the predictor. If disabled, the feature is set to 0 so it has no effect on the prediction. The configuration of features that has been found to work best is enabled by default.


5surim (commented):
Hi @naomirehman1008 

Please check my comments.

Surim

PR #345, opened 2025-07-02
Remove ASLR adjust to execute

Kiminje:
`setarch x86_64 -R`
commented out codes for users who don't want to off ASLR globally.


hlitz (changes requested):
i think we need to disable ASLR for Slurm reasons. please let us know why you need this.

PR #354, opened 2025-07-23
Extract runtime register values from PIN and propagate to Scarab Op structure

Kiminje:
This commit implements a complete pipeline to capture actual register values 
during instruction execution and make them available for dependency analysis 
in Scarab.
[To do]
- ~~PIN extraction~~
- ~~Feed Pin to Scarab Op~~
- ~~Proper register renaming support~~

Key Changes:
- Add src_vals/dst_vals arrays to ctype_pin_inst to store runtime register values
- Add val field to Reg_Info structure to track register values in Op structure  
- Implement register value extraction using PIN's IARG_REG_VALUE at:
  * IPOINT_BEFORE for source register values
  * IPOINT_AFTER for destination register values
- Add is_scarab_compatible_register() filter to only extract values from
  architectural registers (GPR, segment, flags) excluding vector registers
- Update register renaming tables to store and propagate register values
- Modify create_compressed_op() to copy extracted values into instruction info

Flow: PIN extracts runtime values → ctype_pin_inst → Trace_Uop → Inst_Info → Op
This enables precise memory-data-dependent branch analysis using actual
register values rather than symbolic dependencies.

1. STATIC PHASE (Instrumentation):
   - PIN discovers instruction
   - Your callback registers analysis functions
   - PIN compiles instrumented version

2. DYNAMIC PHASE (Every execution):
   - PIN intercepts instruction
   - Executes IPOINT_BEFORE analysis
   - Executes original instruction  
   - Executes IPOINT_AFTER analysis
   - Continues to next instruction

3. FEATURE EXTRACTION:
   - Values captured at precise timing points
   - Stored in global variables temporarily
   - Copied to permanent structures
   - Sent to Scarab for simulation


yzhao321 (changes requested):

hlitz (changes requested):

src/pin/pin_lib/decoder.cc
```diff
@@ -49,6 +49,7 @@ static ctype_pin_inst  tmp_inst_info;
 // Globals used for communication between analysis functions
 uint32_t       glb_opcode, glb_actually_taken;
 deque<ADDRINT> glb_ld_vaddrs, glb_st_vaddrs;
+deque<PIN_REGISTER> glb_src_vector_vals, glb_dst_vector_vals;
```
hlitz: should we make this a pair<REG_ID, REG_VALUE> ?

this would allow us to do checks that the value mathes the actual register id. I want to ensure that we e.g. don't assign the value of a flag dest reg to a generic dst reg and to validate that the reg IDs match etc.

src/pin/pin_lib/decoder.cc
```diff
@@ -305,6 +342,28 @@ void get_branch_dir(bool taken) {
...
+void get_dst_vector_vals(CONTEXT* ctxt, ADDRINT reg_id) {
+  PIN_REGISTER reg_val;
+  PIN_GetContextRegval(ctxt, (REG)reg_id, (UINT8*)&reg_val);
+  glb_dst_vector_vals.push_back(reg_val);
+}
+
+void fill_register_values(ctype_pin_inst* inst, bool is_dst, const deque<PIN_REGISTER>& global_vals, int reg_count) {
+  for (int i = 0; i < reg_count; ++i) {
+    if (is_dst) {
+      inst->dests[i] = global_vals[i].qword[0];
```
hlitz: inst->dests[i] would also have to be a pair<id,val>

src/pin/pin_lib/uop_generator.c
```diff
@@ -76,6 +76,8 @@ struct Trace_Uop_struct {
   Reg_Info dests[MAX_DESTS];
 
   // Dynamic (Runtime) Info
+  uint64_t src_val[MAX_SRCS];
```
hlitz: consider moving src_val into struct Reg_Info, again to ensure that the value is part of the same struct as the ID.

src/ctype_pin_inst.h
```diff
@@ -104,6 +127,8 @@ typedef struct ctype_pin_inst_struct {
 
   /* dynamic information */
 
+  uint64_t srcs[MAX_SRCS];
```
hlitz: maybe we need to define Reg_Info here?

src/map_rename.c
```diff
@@ -733,14 +734,22 @@ void reg_table_entry_consume(struct reg_table_entry *entry, Op *op) {
...
   }
   ASSERT(map_data->proc_id, entry->reg_state == REG_TABLE_ENTRY_STATE_ALLOC);
 
+  entry->reg_val = op->dst_val[dst_reg_idx];
   entry->reg_state = REG_TABLE_ENTRY_STATE_PRODUCED;
   entry->produced_cycle = cycle_count;
+
+  if (!op->off_path) {
+    int arch_id = op->dst_reg_id[dst_reg_idx][REG_TABLE_TYPE_ARCHITECTURAL];
+    map_data->arch_last_val[arch_id] = op->dst_val[dst_reg_idx];
```
hlitz: assert(op->dst.reg_id==arch_id) ?

src/pin/pin_lib/uop_generator.c
```diff
@@ -924,6 +932,12 @@ void convert_dyn_uop(uns8 proc_id, Inst_Info* info, ctype_pin_inst* pi, Trace_Uo
   trace_uop->inst_uid = pi->inst_uid;
   trace_uop->va = 0;
   trace_uop->mem_size = 0;
+  for (uns ii = 0; ii < pi->num_src_regs; ii++) {
+    trace_uop->src_val[ii] = pi->srcs[ii];
```
hlitz: somewhere in uop_dcoder assert(arch_id == srcs[].reg_id);

src/op.h
```diff
@@ -234,6 +234,11 @@ struct Op_struct {
   // }}}
   int bp_confidence;
 
+  // {{{ source and destination values
+  uint64_t src_val[MAX_SRCS];
+  uint64_t dst_val[MAX_DESTS];
+  // }}}
+
   // {{{ register renaming
   uns16 src_reg_id[MAX_SRCS][REG_TABLE_TYPE_NUM];        // the reg id of the source reg file entries
```
hlitz, line 243: use a struct of uns16 and uint64 to combine id and val

src/pin/pin_lib/decoder.cc
hlitz: can we also extract the size of the reg? 8/16/32/64 bit?

we can then also later assert that the size == to decoded_size

hlitz (changes requested):
looks pretty good. all the comments are essentially 1 issue.

please try to validate that the captured values reflect the real values

src/map_rename.c
moonseoppkim: Done [after split id vs after split id]

src/op.h
moonseoppkim: src_reg_id, dst_reg_id, and prev_dst_reg_id are used over 100 times in src/map_rename.c. 

They appear so frequently that it might be a good idea to refactor this file. 
Should we handle this refactoring now, or would it be better to defer it to a future task?

src/pin/pin_lib/decoder.cc
moonseoppkim: Done

src/pin/pin_lib/decoder.cc
moonseoppkim: Done

src/ctype_pin_inst.h
moonseoppkim: Done

src/pin/pin_lib/uop_generator.c
moonseoppkim: Done

src/pin/pin_lib/uop_generator.c
moonseoppkim: generate_uops() splits one PIN inst into multiple uops and injects synthesized temp regs (REG_TMP, REG_RSP, etc.) that aren't in "pi"'s list, so "arch_id == srcs[i].id" breaks on split insts. 

```
debugging log
[SPLIT] pc=0x5555555bed0a uid=23
  BEFORE(pi)  src(0): | dst(0): 
  AFTER(uop)  src(2): RSP RCX | dst(0): 

static uns generate_uops(uns8 proc_id, ctype_pin_inst* pi, Trace_Uop** trace_uop) {
...
  /* Operate */
  if (has_alu) {
    if (has_push || has_pop) {  // ALU op only changes the stack pointer in stack insts
      add_t_uop_src_reg(uop, REG_RSP);<<<<
      add_t_uop_dest_reg(uop, REG_RSP); <<<<
    } else if (!is_rep_st) {
      for (uns i = 0; i < pi->num_ld; ++i) {
        add_t_uop_src_reg(uop, REG_TMP0 + i); <<<<
      }
    }

...
  /* Store */
  if (has_store) {
    for (uns i = 0; i < pi->num_st; ++i) {
      Trace_Uop* uop = trace_uop[idx];
      idx += 1;
      clear_t_uop(uop);

      uop->mem_type = MEM_ST;
      uop->op_type = (pi->is_fp || pi->is_simd) ? OP_FST : OP_IST;
      uop->mem_size = pi->st_size;
      uop->store_seq_num = i;

      if (pi->is_call) {
        // only storing (invisible) EIP on calls
      } else if (!has_alu || has_pop || has_push || is_rep_st) {
        for (uns i = 0; i < pi->num_ld; ++i) {
          add_t_uop_src_reg(uop, REG_TMP0 + i); <<<<
        }
      } else if (has_alu && !has_push && !has_pop) {
        add_t_uop_src_reg(uop, REG_TMP2); <<<<<
      }


static void add_rep_uops(ctype_pin_inst* pi, Trace_Uop** trace_uop, uns* idx) {
...
  if (add_rsi_add) {
...
    add_t_uop_src_reg(uop, REG_RSI);
    add_t_uop_dest_reg(uop, REG_RSI);
  }

  if (add_rdi_add) {
    add_t_uop_src_reg(uop, REG_RDI);
    add_t_uop_dest_reg(uop, REG_RDI);
  }

  if (pi->is_repeat) {
    add_t_uop_src_reg(uop, REG_RCX);
    add_t_uop_dest_reg(uop, REG_RCX);
...
    add_t_uop_src_reg(uop, REG_ZPS);
  }
}
```

src/op.h
hlitz: we can leave it for now

src/pin/pin_lib/uop_generator.c
```diff
@@ -924,6 +930,22 @@ void convert_dyn_uop(uns8 proc_id, Inst_Info* info, ctype_pin_inst* pi, Trace_Uo
...
+      if (pi->src_regs[j] == info->srcs[ii].id) {
+        trace_uop->srcs[ii].val = pi->srcs[j].val;
+        break;
+      }
+  }
+  for (uns ii = 0; ii < info->table_info.num_dest_regs; ii++) {
+    trace_uop->dests[ii].val = 0;
+    for (uns j = 0; j < pi->num_dst_regs; j++)
+      if (pi->dst_regs[j] == info->dests[ii].id) {
+        trace_uop->dests[ii].val = pi->dests[j].val;
```
hlitz, line 947: can we assert that every ii'th val is exactly set once?

src/pin/pin_lib/uop_generator.c
```diff
@@ -400,6 +400,12 @@ void uop_generator_get_uop(uns proc_id, Op* op, ctype_pin_inst* inst) {
   /* multi path support */
 
   /* execute op */
+  for (uns ii = 0; ii < trace_uop->num_src_regs; ii++) {
+    op->src_val[ii] = trace_uop->srcs[ii].val;
```
hlitz, line 404: where are we setting src_reg_id ? also do we store the size somewhere?

Wherever we decode size and reg id we should compare whether it matches srcs[ii].id/size

hlitz (changes requested):
looks pretty good

glad that the test worked fine

src/map.h
```diff
@@ -61,6 +61,10 @@ typedef struct Map_Data_struct {
 
   /* register files for INT/FP with arch/physical tables */
   Reg_File* reg_file[REG_FILE_REG_TYPE_NUM];
+
+  uns64 arch_last_val[NUM_REG_IDS];
+  uns64 arch_last_uid[NUM_REG_IDS];
+  Flag arch_last_valid[NUM_REG_IDS];
 } Map_Data;
```
yzhao321, line 64: 1. What is the motivation of having these three fields? I saw they are used in the function of "reg_value_read" but the function is not called;
2. Why are they put in Map_Data instead of reg_table_entry of arch table? I think it is better that Map_Data and map.c is only for dependency.

src/map_rename.c
```diff
@@ -733,14 +734,23 @@ void reg_table_entry_consume(struct reg_table_entry *entry, Op *op) {
...
 
+  entry->reg_val = op->dst_val[dst_reg_idx];
   entry->reg_state = REG_TABLE_ENTRY_STATE_PRODUCED;
   entry->produced_cycle = cycle_count;
+
+  if (!op->off_path) {
+    int arch_id = op->dst_reg_id[dst_reg_idx][REG_TABLE_TYPE_ARCHITECTURAL];
+    ASSERT(map_data->proc_id, op->inst_info->dests[dst_reg_idx].id == arch_id);
+    map_data->arch_last_val[arch_id] = op->dst_val[dst_reg_idx];
+    map_data->arch_last_uid[arch_id] = op->inst_uid;
```
yzhao321: Is there a scenario that the arch table entry is not the current physical reg? e.g.,  I1 rename r1 and get ptag1, I2 rename r1 and get ptag2, then I1 produce ptag1. Then the r1 currently is for ptag2 not for ptag1.

src/map_rename.c
```diff
@@ -1927,3 +1937,12 @@ void reg_file_commit(Op *op) {
...
 }
+
+Flag reg_value_read(int arch_id, uns64 *val_out, uns64 *uid_out) {
+  if (!map_data->arch_last_valid[arch_id])
+    return FALSE;
+
+  *val_out = map_data->arch_last_val[arch_id];
+  *uid_out = map_data->arch_last_uid[arch_id];
+  return TRUE;
+}
```
yzhao321: Can you introduce a future case that will use this func and its corresponding val_out and uid_out?

src/map_rename.h
```diff
@@ -89,6 +89,9 @@ struct reg_table_entry {
   // type info
   int reg_type;        // INT or VEC
   int reg_table_type;  // arch, physical, or virtual
+  // register value
+  uint64_t reg_val;
+  uns64 produced_uid;
```
yzhao321, line 95: Can arch_entry and phy_entry share these same field? Then the propogation during producing can be written by arch_instance->val = phy_instance->val

src/op.h
```diff
@@ -234,6 +234,11 @@ struct Op_struct {
   // }}}
   int bp_confidence;
 
+  // {{{ source and destination values
+  uint64_t src_val[MAX_SRCS];
+  uint64_t dst_val[MAX_DESTS];
+  // }}}
```
yzhao321, line 240: These values are pre-know during the injection but cannot be consumed until producing right?

src/op.h
yzhao321, line 246: Are there any drawbacks to put these fileds here?
1. For src_reg_id, dst_reg_id, they are the properties of ops and I believed in the hardware they are also stored in an op. Then it use these ids to query the map table as far as i am concerned?
2. For prev_dst_reg_id, it can either be put in here (should be ROB but there is not rob entry instance) or registers. In the paper we assumed that they are put in the ROB then we can get the higher bit reduction for adapting new tech. But storing in the ROB has the drawback of requring extra physical accessing when doing the register releasing.

yzhao321 (changes requested):
Hi @moonseoppkim @hlitz Thanks for your commits and revies. I left some questions about this PR.

src/map.h
hlitz: I think yinyuan is right. If we write the values into the physical register, we can use the SRT to lookup the current architectural to physical mapping and then read the value from the physical reg file.

arch_last_val should hence be redundant. I think we can remove it to avoid inconsistencies.

src/map_rename.c
```diff
@@ -733,14 +734,23 @@ void reg_table_entry_consume(struct reg_table_entry *entry, Op *op) {
...
 
 /* update the register state to indicate the value is produced during execution*/
-void reg_table_entry_produce(struct reg_table_entry *entry, Op *op) {
+void reg_table_entry_produce(struct reg_table_entry *entry, Op *op, uns dst_reg_idx) {
   if (op->move_eliminated) {
     return;
   }
   ASSERT(map_data->proc_id, entry->reg_state == REG_TABLE_ENTRY_STATE_ALLOC);
 
+  entry->reg_val = op->dst_val[dst_reg_idx];
```
hlitz, line 744: can we add entry->reg_val_on_path which states whether the reg val was written by an on or off-path instruction?

src/map_rename.c
hlitz: If we remove arch_last_val and rely on phys reg, this issue should be resolved.

src/pin/pin_lib/uop_generator.c
moonseoppkim: Done

* comment 
2. Distribute PIN-collected values to split uops

   Expected found == 1, but uop can be explicitly splited into duplicate same reg ids, like below debugging log

     [VAL-SRC] uid=205 found=2 
        pi_reg=8(RSP) num_uop=2
        splited uop[0] srcs: 8(RSP)
        splited uop[1] srcs: 8(RSP) << inserted from “generate_uops()” by “add_t_uop_src_reg()”

   found >= 2 is valid; only found == 0 (value loss) is a bug.

src/pin/pin_lib/uop_generator.c
moonseoppkim: Done

I've added more assertion at decoding ("fill_register_values()")

src/map.h
moonseoppkim: 1. I'm planning to implement a new data-dependent branch predictor which leverages arch register values for predictions.

2. remove these 3 variables

src/map_rename.c
moonseoppkim: Future data-dependent branch predictor will utilize these values.

src/op.h
moonseoppkim: Yes, you are correct.

src/map_rename.c
moonseoppkim: Great catch, I can fix with a uid guard so an older definition can't overwrite a younger value,
If you have a better idea, please let me know!

src/map_rename.c
hlitz: I think it is good to have this API but we should lookup the arch_id in SRT and then retrun value from phys reg.

may

moonseoppkim: Done

src/map_rename.c
moonseoppkim: Done

src/map.h
moonseoppkim: Done

src/map_rename.c
moonseoppkim: Done

src/map_rename.h
moonseoppkim: Produce is physical only by design("reg_file_produce_dst" asserts table_type > ARCHITECTURAL), 
and the arch table's ops (reg_table_ops_arch) don't even register a .produce handler, so arch entries never hold values. 

```
struct reg_table_ops reg_table_ops = {
    .init    = reg_table_init,
    .read    = reg_table_read,
    .alloc   = reg_table_alloc,
    .free    = reg_table_free,
    .consume = reg_table_consume,
    .produce = reg_table_produce,
};

struct reg_table_ops reg_table_ops_arch = {
    .init = reg_table_arch_init,
};
```

Hence I read via child_reg_id -> physical entry rather than propagating val into the arch entry.

src/op.h
moonseoppkim: Agreed, src_reg_id and dst_reg_id belong in the op structure. On whether relocating prev_dst_reg_id would be beneficial.

If you really want to adjust the code structure, let's discuss more at the separated tab.

src/pin/pin_lib/message_queue_interface_lib.h
```diff
@@ -39,7 +39,7 @@ extern "C" {
 #include <vector>
 #include "pin_scarab_common_lib.h"
 
-#define MAX_PACKET_SIZE (1 << 12)
+#define MAX_PACKET_SIZE (1 << 14)
```
moonseoppkim, line 42: Message Queue Assertion Fired: Scarab does not currently support sending messages larger than MAX_PACKET_SIZE.

src/pin/pin_lib/message_queue_interface_lib.cc
```diff
@@ -24,7 +24,7 @@
 
 extern "C" {}
 
-#define RECEIVE_BUFFER_MAX_SIZE (0x01 << 12)
+#define RECEIVE_BUFFER_MAX_SIZE (0x01 << 14)
```
moonseoppkim, line 27: Message Queue Assertion Fired: Need to allocation more space in the send buffer

src/map_rename.c
```diff
@@ -1927,3 +1933,37 @@ void reg_file_commit(Op *op) {
...
+
+Flag reg_value_read(int arch_id, uns64 *val_out, uns64 *uid_out) {
+  int reg_type = reg_file_get_reg_type(arch_id);
+  struct reg_table *srt = map_data->reg_file[reg_type]->reg_table[REG_TABLE_TYPE_ARCHITECTURAL];
+  struct reg_table *ptab = map_data->reg_file[reg_type]->reg_table[REG_TABLE_TYPE_PHYSICAL];
+
+  int ptag = srt->entries[arch_id].child_reg_id;
+  int guard = ptab->size;
+
+  while (ptag != REG_TABLE_REG_ID_INVALID && guard-- > 0) {
```
hlitz: If I understand it returns the oldest, produced, on-path physical value for a given arch_id. This is not realistic to implement.

First we do not know whether values are on-path. Second, doing a series of lookups based on prev_id is difficult to implement as well.

I also do not know whether it would be useful. How are you planning to use such an API?

I would have suggested a read API that does a single SRT lookup and then return the value and on-path flag and whether the value has been produced. But we can also push back the API to later when we actually know what we need

src/map_rename.c
```diff
@@ -1927,3 +1933,37 @@ void reg_file_commit(Op *op) {
          REG_RENAMING_SCHEME >= REG_RENAMING_SCHEME_INFINITE && REG_RENAMING_SCHEME < REG_RENAMING_SCHEME_NUM);
   reg_renaming_scheme_func_table[REG_RENAMING_SCHEME].commit(op);
 }
+
+Flag reg_value_read(int arch_id, uns64 *val_out, uns64 *uid_out) {
+  int reg_type = reg_file_get_reg_type(arch_id);
+  struct reg_table *srt = map_data->reg_file[reg_type]->reg_table[REG_TABLE_TYPE_ARCHITECTURAL];
+  struct reg_table *ptab = map_data->reg_file[reg_type]->reg_table[REG_TABLE_TYPE_PHYSICAL];
+
+  int ptag = srt->entries[arch_id].child_reg_id;
```
hlitz: I think we need to do multiple lookups if virtual ptag is supported. @yzhao321 to confirm. 

I don't think we need to support virtual ptag right now but we should add an assert taht reg_value_read does not support vtag.

hlitz (changes requested):

src/pin/pin_lib/decoder.cc
```diff
@@ -188,10 +211,41 @@ void insert_analysis_functions(ctype_pin_inst* info, const INS& ins) {
...
                  IARG_INST_PTR, IARG_END);
 }
 
+void create_compressed_op_after(ADDRINT iaddr) {
+  if (!fast_forward_count) {
+    assert(inst_info_storage.count(iaddr) == 1);
+    filled_inst_info = inst_info_storage[iaddr];
+    fill_register_values(filled_inst_info, true, glb_dst_vector_vals, filled_inst_info->num_dst_regs);
+  }
+}
```
yzhao321, line 255: NIT: 
if (fast_forward_count) {
   early return;
}

do logic

src/pin/pin_lib/decoder.cc
```diff
@@ -305,6 +362,54 @@ void get_branch_dir(bool taken) {
...
+      ASSERTX(global_vals[i].id == inst->dst_regs[i]);
+      inst->dests[i].id = global_vals[i].id;
+      inst->dests[i].val = global_vals[i].val;
+      inst->dests[i].size = global_vals[i].size;
+    } else {
+      ASSERTX(global_vals[i].id == inst->src_regs[i]);
+      inst->srcs[i].id = global_vals[i].id;
+      inst->srcs[i].val = global_vals[i].val;
+      inst->srcs[i].size = global_vals[i].size;
+    }
```
yzhao321, line 413: NIT:
if (is_dst) {
 ...
 continue;
}

... do other logic

remove else helps reduce indentation to reduce the pressure of reading code.

src/pin/pin_lib/decoder.cc
```diff
@@ -188,10 +211,41 @@ void insert_analysis_functions(ctype_pin_inst* info, const INS& ins) {
...
+        uint8_t scarab_id = info->dst_regs[i];
+        auto it = reg_scarab_to_pin_map.find(scarab_id);
+        REG pin_reg = (it != reg_scarab_to_pin_map.end()) ? it->second : REG_INVALID_;
+        INS_InsertCall(ins, IPOINT_AFTER, (AFUNPTR)get_dst_vector_vals, IARG_CONST_CONTEXT, IARG_ADDRINT,
+                       (ADDRINT)pin_reg, IARG_ADDRINT, (ADDRINT)scarab_id, IARG_END);
+      }
+      INS_InsertCall(ins, IPOINT_AFTER, (AFUNPTR)create_compressed_op_after, IARG_INST_PTR, IARG_END);
+    }
+  }
+
```
yzhao321, line 242: Should we add some comments here? this is a very long function (75+ lines) without comments and I cannot understant them.

src/map_rename.h
```diff
@@ -89,6 +89,10 @@ struct reg_table_entry {
...
+  Flag reg_val_on_path;
```
yzhao321: 1. A comment should be added to explain the meaning of this.
2. why is it not reg_val_off_path to align with the same default of op/entry->off_path?

src/map_rename.c
```diff
@@ -1927,3 +1933,37 @@ void reg_file_commit(Op *op) {
          REG_RENAMING_SCHEME >= REG_RENAMING_SCHEME_INFINITE && REG_RENAMING_SCHEME < REG_RENAMING_SCHEME_NUM);
   reg_renaming_scheme_func_table[REG_RENAMING_SCHEME].commit(op);
 }
+
+Flag reg_value_read(int arch_id, uns64 *val_out, uns64 *uid_out) {
+  int reg_type = reg_file_get_reg_type(arch_id);
```
yzhao321: I agree with that this API should be removed in this PR

src/map_rename.c
yzhao321: Multiple lookup is needed for virtual table.

yzhao321 (changes requested):

src/map_rename.h
moonseoppkim: 1. Done
2. Yes, you are right. I've removed the redundant new variable and we can use the op/entry->off_path instead.

src/pin/pin_lib/decoder.cc
moonseoppkim: Done

src/pin/pin_lib/decoder.cc
moonseoppkim: Done

src/pin/pin_lib/decoder.cc
moonseoppkim: Done

src/map_rename.c
moonseoppkim: I got rid of this API, and get back later with new implementation.

src/map_rename.c
moonseoppkim: Okay, makes sense. 
I've dropped the API from this PR, and will get back to you with a more concrete idea of the new branch predictor.

* On the intended use: 
The plan is a hashed-perceptron predictor integrated as a new component inside TAGE, taking all GPR values as input features(similar to my earlier offline experiments) + runahead distance + etc.

moonseoppkim: @yzhao321 

If I implement the read func using the existing map_data reg_file/reg_table path (a single SRT lookup), there's a problem, hot registers would return "not produced(invalid)" most of the time, since their youngest mapping is almost always in-flight. 

e.g.
[t0] add r15, ...  renamed -> P12 allocated (ALLOC), SRT[r15] = P12, value empty
[t1] branch fetched -> predictor reads SRT[r15] -> P12 is ALLOC -> "not produced"
[t2] add finally executes -> P12 gets its value, but the prediction already happened

The separated register mirror table I added earlier would solve this, single table lookup that always returns the last produced value.

hlitz (approved):
LGTM

src/pin/pin_lib/decoder.cc
```diff
@@ -305,6 +370,55 @@ void get_branch_dir(bool taken) {
...
+  if (REG_valid(reg)) {
+    PIN_GetContextRegval(ctxt, reg, (UINT8*)&reg_val);
+    val = reg_val.qword[0];
+    size = (uint8_t)(REG_Size(reg) * 8);
+  }
+  glb_dst_vector_vals.push_back({(uint16_t)scarab_id, val, size});
+}
+
+static inline bool is_arch_gpr(uint8_t reg_id) {
+  return reg_id >= SCARAB_REG_RAX && reg_id <= SCARAB_REG_R15;
```
yzhao321, line 400: Sorry one last question: What is the motivation of introducing SCARAB_REG? Why can't we directly use REG_RAX?

yzhao321 (changes requested):
Thanks for your update. I have one more question about SCARAB_REG. Thanks for your time.

src/pin/pin_lib/decoder.cc
moonseoppkim: Long story to short : 
Pin side -> SCARAB_REG_RAX ~ SCARAB_REG_R15 (1~16)
                   REG_RAX ~ REG_R15 (10, 7, 9, 8, ...)

Scarab side -> REG_RAX ~ REG_R15 (1 ~ 16), like pin_side SCARAB_REG_*

************************************************************************


The three libraries in the pipeline each use their own register numbering, 
so the same architectural register has three different ids:

```
under pin side code, (src/pin/pin_lib/decoder.cc)
────────────────────────────────────────────────────────────────
 ARCH REG │      XED (id)      │      PIN (id)      │ SCARAB (id)
────────────────────────────────────────────────────────────────
 RAX      │ XED_REG_RAX   =  98│ REG_RAX      =  10 │ SCARAB_REG_RAX  =  1
 RBX      │ XED_REG_RBX   = 101│ REG_RBX      =   7 │ SCARAB_REG_RBX  =  2
 RCX      │ XED_REG_RCX   =  99│ REG_RCX      =   9 │ SCARAB_REG_RCX  =  3
 RDX      │ XED_REG_RDX   = 100│ REG_RDX      =   8 │ SCARAB_REG_RDX  =  4
 RBP      │ XED_REG_RBP   = 103│ REG_RBP      =   5 │ SCARAB_REG_RBP  =  5
 RSI      │ XED_REG_RSI   = 104│ REG_RSI      =   4 │ SCARAB_REG_RSI  =  6
 RDI      │ XED_REG_RDI   = 105│ REG_RDI      =   3 │ SCARAB_REG_RDI  =  7
 RSP      │ XED_REG_RSP   = 102│ REG_RSP      =   6 │ SCARAB_REG_RSP  =  8
 R8       │ XED_REG_R8    = 106│ REG_R8       =  11 │ SCARAB_REG_R8   =  9
 R9       │ XED_REG_R9    = 107│ REG_R9       =  12 │ SCARAB_REG_R9   = 10
 R10      │ XED_REG_R10   = 108│ REG_R10      =  13 │ SCARAB_REG_R10  = 11
 R11      │ XED_REG_R11   = 109│ REG_R11      =  14 │ SCARAB_REG_R11  = 12
 R12      │ XED_REG_R12   = 110│ REG_R12      =  15 │ SCARAB_REG_R12  = 13
 R13      │ XED_REG_R13   = 111│ REG_R13      =  16 │ SCARAB_REG_R13  = 14
 R14      │ XED_REG_R14   = 112│ REG_R14      =  17 │ SCARAB_REG_R14  = 15
 R15      │ XED_REG_R15   = 113│ REG_R15      =  18 │ SCARAB_REG_R15  = 16
────────────────────────────────────────────────────────────────
```

I needed to use "SCARAB_REG_*" series in order to avoid naming conflicts with PIN reg_id under pin side.
But the same enumeration (REG_RAX ~ REG_R15, defined in isa/isa.h) is used directly on the Scarab side (e.g., map_rename.c), where REG_RAX = 1.

```
scarab/src/isa/isa.h
#define REG(x) REG_##x,
typedef enum Reg_Id_struct {
#include "isa/x86_regs.def"
  NUM_REGS
} Reg_Id;
#undef REG

// map_rename.c just includes it:
scarab/src/map_rename.c:40
#include "isa/isa.h"
```

yzhao321 (approved):
LGTM

src/map_rename.c
moonseoppkim: @yzhao321 

Do you agree we should revive the separated last_value structure again in order to guarantee one-shot register value table lookup??

yzhao321: Hi @moonseoppkim  I missed this comment thanks for your reminding. I need more context about the work. I cannot get the necessity of reading vals from the arch table instead of the physical table. I think the register values in the arch table are always not produced makes sense as the arch table is to remove the false dependencies. Why cannot directly get the value by physical id?

moonseoppkim: @yzhao321 

The consumer here is the branch predictor at fetch/predict time(RAX~R15) will be the input feature for hashed perceptron inside the TAGE, not a dependent instruction in rename. 

"The predictor only knows the arch id", so it has no physical id in hand. The only arch->phys mapping is the SRT.

The problem: the SRT always points to the youngest mapping, which for hot registers is almost always still in-flight (ALLOC) at prediction time. What the predictor needs is the last produced value, and reaching that through the physical table would require walking the prev_id chain, which Heiner already mentioned as unrealistic.

## Proposal1, separated register mirror table structure
So the proposal is not to read values from the arch table itself. It's a separate arch-id-indexed last_value mirror table, updated at produce time, giving a one-shot lookup of the most recently produced value per arch register.


## Proposal2, another approach without mirror table
An alternative without a mirror table would be scanning the physical table at every prediction: 
for each arch reg (RAX–R15), walk all ~ PRF entries, 
"filter state >= PRODUCED && parent_reg_id == arch_id", and pick the max-produced_uid entry. 
This works functionally, 

but: I don't think it's a realistic implementation.

PR #450, opened 2026-03-18
BTB: Add block btb mechanism

estodi:
This PR adds baseline Block BTB (B-BTB) mechanism. In this baseline implementation, cf ops that overflow an entry due to the lack of branch slots within a block are simply not stored in the entry, leading to performance degradation.

- `baseline`: base of this PR
- `generic`: this PR with `--btb_mech generic`
- `brslots_X`: this PR with `--btb_mech block --btb_num_brslot X`

Workload | baseline (IPC) | generic (IPC) | brslots_1 (IPC) | brslots_2 (IPC) | brslots_3 (IPC) | brslots_4 (IPC) | generic speedup vs baseline (%) | brslots_1 speedup vs baseline (%) | brslots_2 speedup vs baseline (%) | brslots_3 speedup vs baseline (%) | brslots_4 speedup vs baseline (%)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
spec2017/rate_int_v2/deepsjeng_r | 2.4509 | 2.4509 | 2.1185 | 2.3851 | 2.4224 | 2.4255 | 0% | -13.56% | -2.68% | -1.16% | -1.03%
spec2017/rate_int_v2/exchange2_r | 3.5081 | 3.5081 | 2.5693 | 3.1325 | 3.2147 | 3.5214 | 0% | -26.76% | -10.71% | -8.36% | 0.38%
spec2017/rate_int_v2/gcc_r | 8.0959 | 8.0959 | 6.9586 | 7.7801 | 7.9088 | 7.9456 | 0% | -14.05% | -3.9% | -2.31% | -1.86%
spec2017/rate_int_v2/gcc_r_2 | 1.8318 | 1.8318 | 1.5787 | 1.7647 | 1.7909 | 1.7984 | 0% | -13.81% | -3.66% | -2.23% | -1.82%
spec2017/rate_int_v2/gcc_r_3 | 1.1296 | 1.1296 | 1.0552 | 1.1082 | 1.1129 | 1.1159 | 0% | -6.58% | -1.89% | -1.48% | -1.21%
spec2017/rate_int_v2/gcc_r_4 | 2.0131 | 2.0131 | 1.7719 | 1.943 | 1.973 | 1.9825 | 0% | -11.98% | -3.48% | -1.99% | -1.52%
spec2017/rate_int_v2/gcc_r_5 | 1.6542 | 1.6542 | 1.3009 | 1.5653 | 1.6109 | 1.6214 | 0% | -21.36% | -5.38% | -2.62% | -1.98%
spec2017/rate_int_v2/leela_r | 1.6739 | 1.6739 | 1.5121 | 1.6345 | 1.6593 | 1.6642 | -0% | -9.67% | -2.36% | -0.88% | -0.58%
spec2017/rate_int_v2/mcf_r | 0.7831 | 0.7832 | 0.7291 | 0.7779 | 0.7784 | 0.779 | 0% | -6.9% | -0.67% | -0.61% | -0.53%
spec2017/rate_int_v2/omnetpp_r | 0.9437 | 0.9437 | 0.8988 | 0.9358 | 0.9385 | 0.9394 | 0% | -4.76% | -0.84% | -0.55% | -0.46%
spec2017/rate_int_v2/perlbench_r | 8.3623 | 8.3623 | 7.2319 | 8.1117 | 8.26 | 8.3227 | 0% | -13.52% | -3% | -1.22% | -0.47%
spec2017/rate_int_v2/perlbench_r_2 | 3.3188 | 3.3188 | 2.6857 | 3.2052 | 3.2842 | 3.2972 | 0% | -19.08% | -3.42% | -1.04% | -0.65%
spec2017/rate_int_v2/perlbench_r_3 | 2.1411 | 2.1411 | 2.0867 | 2.1297 | 2.1334 | 2.1345 | 0% | -2.54% | -0.54% | -0.36% | -0.31%
spec2017/rate_int_v2/x264_r | 13.4785 | 13.4785 | 13.0434 | 13.4439 | 13.4711 | 13.4667 | 0% | -3.23% | -0.26% | -0.06% | -0.09%
spec2017/rate_int_v2/x264_r_2 | 4.6302 | 4.6302 | 4.4985 | 4.6215 | 4.6271 | 4.6271 | 0% | -2.85% | -0.19% | -0.07% | -0.07%
spec2017/rate_int_v2/x264_r_3 | 4.5439 | 4.5439 | 4.3912 | 4.5261 | 4.537 | 4.5326 | 0% | -3.36% | -0.39% | -0.15% | -0.25%
spec2017/rate_int_v2/xalancbmk_r | 1.0649 | 1.0649 | 0.9411 | 0.9743 | 1.0383 | 1.0597 | 0% | -11.63% | -8.51% | -2.5% | -0.49%
spec2017/rate_int_v2/xz_r | 5.0183 | 5.0183 | 4.2881 | 5.0387 | 5.0114 | 5.0161 | 0% | -14.55% | 0.41% | -0.14% | -0.04%
spec2017/rate_int_v2/xz_r_2 | 2.5605 | 2.5605 | 2.0101 | 2.5748 | 2.5583 | 2.5611 | 0% | -21.49% | 0.56% | -0.09% | 0.02%
spec2017/rate_int_v2/xz_r_3 | 1.6296 | 1.6296 | 1.473 | 1.6203 | 1.6264 | 1.6282 | 0% | -9.61% | -0.57% | -0.19% | -0.08%
Avg | 3.5416 | 3.5416 | 3.1571 | 3.4637 | 3.4978 | 3.522 | 0% | -10.86% | -2.2% | -1.24% | -0.55%


src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+  return;
 }
 
 /**************************************************************************************/
 /* bp_btb_block_update: */
 
 void bp_btb_block_update(Bp_Data* bp_data, Op* op) {
-  bp_btb_gen_update(bp_data, op);
+  ASSERT(bp_data->proc_id, bp_data->proc_id == op->proc_id);
+  if (!op->off_path) {
```
hlitz: for the other BTBs don't we check whether UPDATE_OFF_PATH == true or something? should do this here as well

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+        // Invalidate the remaining slots
+        for (uns ii = 1; ii < BTB_NUM_BRSLOT; ii++)
+          br_slots[ii].valid = FALSE;
+      } else {
+        uns insert_pos = BTB_NUM_BRSLOT;
+        for (uns ii = 0; ii < BTB_NUM_BRSLOT; ii++) {
+          if (br_slots[ii].valid) {
+            if (br_slots[ii].addr < op->inst_info->addr) {
+              continue;
+            } else if (br_slots[ii].addr == op->inst_info->addr) {
```
hlitz: is this for self-modifying code? or why should br_slots[ii].addr != op->inst_info->addr) ?

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+              br_slots[ii] = br_slot;
+              break;
+            } else {
+              // This assertion holds only when there is no self-modification code (e.g. SPEC 2017)
+              ASSERT(bp_data->proc_id,
```
hlitz: we need to be able to run other applications as well. We can't have an assert that fails for non-spec.

the clean way would be to look at the encoding_is_new flag but it is not supported by PT yet

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+                     op->inst_info->table_info.cf_type == CF_CBR || op->inst_info->table_info.cf_type == CF_REP);
+              if (op->inst_info->table_info.cf_type == CF_CBR || op->inst_info->table_info.cf_type == CF_REP) {
+                // If this op is NOT always-taken, it needs to be inserted, not just appended
+                insert_pos = ii;
+              } else {
```
hlitz, line 652: Because of the assertion, this case can never happen right? Is it needed for SMC ?

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+            // br_slots[ii] does not store a valid op yet
+            br_slots[ii] = br_slot;
+            break;
+          }
+        }
+
+        if (insert_pos < BTB_NUM_BRSLOT) {
+          // Naive replacement policy: op with the largest addr will be discarded
+          for (uns ii = BTB_NUM_BRSLOT - 1; ii > insert_pos; ii--) {
+            if (br_slots[ii - 1].valid) {
```
hlitz: could we break as soon as we find an invalid entry to avoid discarding largest addr?

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+              br_slots[ii] = br_slots[ii - 1];
+            }
+          }
+          br_slots[insert_pos] = br_slot;
+        }
+      }
+
+      // FIXME: the exceptions to this assert are really about x86 vs Alpha
```
hlitz: i dont understand. Please elaborate why Alpha which is an ISA we do not support

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+  }
+
+  Addr btb_index_addr = 0;
+  // Actual BTB does not require this because the index addr to look up BTB is given,
+  // but Scarab needs to memorize the previous target as it looks up BTB without knowing the index addr.
+  if (bp_data->prev_cf_btb_index_addr == 0) {
+    // Only for the very first access
+    btb_index_addr = ft_get_ft_info(op->parent_FT).static_info.start;
+  } else {
+    if (bp_data->prev_cf_pred) {
```
hlitz: is this taken/not taken?

in other places this is called dir. To make it more readable, do:

bp_data->dir == TAKEN (I think there is a TAKEN constant defined somehwere)

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+    bp_data->btb = primary_bp->btb;
 }
 
 /**************************************************************************************/
 /* bp_btb_block_pred: */
 
 void bp_btb_block_pred(Bp_Data* bp_data, Op* op) {
-  bp_btb_gen_pred(bp_data, op);
+  Btb_Pred_Info* bpi = op->btb_pred_info;
+
```
hlitz, line 555: can we handle non-cf? should we assert that op->cf is true?

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+    // Only for the very first access
+    btb_index_addr = ft_get_ft_info(op->parent_FT).static_info.start;
+  } else {
+    if (bp_data->prev_cf_pred) {
+      btb_index_addr = bp_data->prev_cf_target;
+    } else {
+      btb_index_addr = bp_data->prev_cf_btb_index_addr;
+    }
+  }
+  uns64 _blk_idx = (op->inst_info->addr - btb_index_addr) / BTB_BLOCK_SIZE;
```
hlitz: because the compiler cannot proof that this is a power of two, it will produce division instruction. change to bit shift explicitly by using LOG_BTB_BLOCK_SIZE

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+  btb_index_addr += BTB_BLOCK_SIZE * _blk_idx;
```
hlitz: actually, you are dividing and then multiplying with the same number, you should mask out bits instead

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+  ASSERT(bp_data->proc_id, btb_index_addr <= op->inst_info->addr);
+
+  // Store index for update
+  op->btb_pred_info->btb_index_addr = btb_index_addr;
```
hlitz, line 581: so if I understand correctly, this computation makes the btb_index_addr block_size-aligned. Why are we doing this again? Don't we need to index the BTB with the actual start of the BBL not the block-aligned address of the BBL?

What if two BBLs are in the same block? Wouldn't they alias?

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+
+  // Prepare for next BTB lookup
+  bp_data->prev_cf_target = op->oracle_info.target;
+  bp_data->prev_cf_btb_index_addr = btb_index_addr;
+
+  Addr btb_line_addr;
+  Blk_Btb_BrSlot* br_slots = (Blk_Btb_BrSlot*)cache_access(bp_data->btb, btb_index_addr, &btb_line_addr, TRUE);
+
+  if (br_slots) {
+    for (uns ii = 0; ii < BTB_NUM_BRSLOT; ii++) {
```
hlitz, line 592: for loop can be simplified.

- define hit=false at beginning
- eliminate if (br_slots[ii].addr < op->inst_info->addr) {
- only set hit=true for the equal case 
- eliminate the returns (breaks/returns are more expensive than just executing a fixed number of loop iterations if the loop iterations are small e.g. 4)
- at the end set btb_main_hit once before the return

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +519,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+                                                 &repl_line_addr);
+        br_slots[0] = br_slot;
+        // Invalidate the remaining slots
+        for (uns ii = 1; ii < BTB_NUM_BRSLOT; ii++)
+          br_slots[ii].valid = FALSE;
+      } else {
+        uns insert_pos = BTB_NUM_BRSLOT;
+        for (uns ii = 0; ii < BTB_NUM_BRSLOT; ii++) {
+          if (br_slots[ii].valid) {
+            if (br_slots[ii].addr < op->inst_info->addr) {
```
hlitz: these two lines can be eliminated

hlitz (changes requested):

src/bp/bp_targ_mech.c
estodi: I believe BTB should work regardless of the cf type of the given op. Currently, BTB pred func is called inside `bp_predict_btb`, which is handling non-cf before calling BTB pred.

hlitz: are you sure that this code will not break for non-cf ops? for isntance btb_pred_info may not have been initizlised

estodi: We can assume `btb_pred_info` is always initialized whenever BTB pred (`bp_data->bp_btb->pred_func`) is called, as it is only called from `bp_predict_btb` that initializes `btb_pred_info` for every op upon entry: https://github.com/litz-lab/scarab_ll/blob/044c8f95e5b121fd013266905add6c5e135a4326/src/bp/bp_targ_mech.c#L293

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +522,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+    btb_index_addr = ft_get_ft_info(op->parent_FT).static_info.start;
+  } else {
+    if (bp_data->prev_cf_pred == TAKEN) {
+      btb_index_addr = bp_data->prev_cf_target;
+    } else {
+      btb_index_addr = bp_data->prev_cf_btb_index_addr;
+    }
+  }
+  // Compute block-size-aligned (fall-through) address from the start of the first block, in case op is far away.
+  btb_index_addr += (op->inst_info->addr - btb_index_addr) & ~(BTB_BLOCK_SIZE - 1);
```
hlitz, line 577: I thought we would do like this (double check correctness):
`
uint64_t btb_block_offset = btb_index_addr & (BTB_BLOCK_SIZE - 1);
uint64_t btb_index_addr_masked = op->inst_info->addr & ~(BTB_BLOCK_SIZE - 1);
btb_index_addr = btb_index_addr_masked | `btb_block_offset;`
`

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +522,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+  bpi->btb_main_hit = FALSE;
+  if (br_slots) {
+    for (uns ii = 0; ii < BTB_NUM_BRSLOT; ii++) {
+      if (br_slots[ii].valid) {
+        if (br_slots[ii].addr == op->inst_info->addr) {
+          bpi->btb_main_hit = TRUE;
+          bpi->btb_main_target = br_slots[ii].target;
+          break;
+        }
+      } else {
```
hlitz: remove this else case (it will not affect correctness and will improve reaability and speed).

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +522,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+  bp_data->prev_cf_btb_index_addr = btb_index_addr;
+
+  Addr btb_line_addr;
+  Blk_Btb_BrSlot* br_slots = (Blk_Btb_BrSlot*)cache_access(bp_data->btb, btb_index_addr, &btb_line_addr, TRUE);
+
+  bpi->btb_main_hit = FALSE;
+  if (br_slots) {
+    for (uns ii = 0; ii < BTB_NUM_BRSLOT; ii++) {
+      if (br_slots[ii].valid) {
+        if (br_slots[ii].addr == op->inst_info->addr) {
```
hlitz: combine both ifs into single statement to reduce indentation (helps readability)

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +522,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+                                                 &repl_line_addr);
+        br_slots[0] = br_slot;
+        // Invalidate the remaining slots
+        for (uns ii = 1; ii < BTB_NUM_BRSLOT; ii++)
+          br_slots[ii].valid = FALSE;
+      } else {
+        uns insert_pos = BTB_NUM_BRSLOT;
+        for (uns ii = 0; ii < BTB_NUM_BRSLOT; ii++) {
+          if (br_slots[ii].valid) {
+            if (br_slots[ii].addr < op->inst_info->addr) {
```
hlitz: these have not been resolved! please fix (remove the if).

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +522,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+              continue;
+            } else if (br_slots[ii].addr == op->inst_info->addr) {
+              br_slots[ii] = br_slot;
```
hlitz, line 642: you haven't answered this. why are we overwriting the entry? self-modifying code?

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +522,162 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+            } else {
+              // If there is no self-modifying code (e.g. SPEC 2017 int), op must be conditional in this case.
+              // Enable this assertion when debugging.
+              // ASSERT(bp_data->proc_id,
+              //        op->inst_info->table_info.cf_type == CF_CBR || op->inst_info->table_info.cf_type == CF_REP);
+              if (op->inst_info->table_info.cf_type == CF_CBR || op->inst_info->table_info.cf_type == CF_REP) {
+                // If this op is NOT always-taken, it needs to be inserted, not just appended.
+                insert_pos = ii;
+              } else {
+                // If this op is always-taken, invalidate the rest as the block ends here.
```
hlitz, line 653: is it even possible that this case happens (ignoring SMC)? shouldn't we have seen that always-taken branch CF before?

src/bp/bp_targ_mech.c
hlitz: you want to remove the term Alpha. it is not supported, just because you copied it does not make it correct.

hlitz (changes requested):

src/bp/bp_targ_mech.c
estodi: The suggested OR approach was not logicaly equivalent to the arithmetic approach.
For example, with `BTB_BLOCK_SIZE` being 64, when btb index addr is `0x7ffff31a41f8` and op addr is `0x7ffff31a4204`, we would expect `0x7ffff31a41f8` as the start address, but OR approach gives `0x7ffff31a4238`.

src/bp/bp_targ_mech.c
estodi: This case applies to SMC and indirect branches.

src/bp/bp_targ_mech.c
estodi: Ignoring SMC, this does not happen.

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +522,158 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+
+  bpi->btb_main_hit = FALSE;
+  if (br_slots) {
+    for (uns ii = 0; ii < BTB_NUM_BRSLOT; ii++) {
+      if (br_slots[ii].valid && br_slots[ii].addr == op->inst_info->addr) {
+        bpi->btb_main_hit = TRUE;
+        bpi->btb_main_target = br_slots[ii].target;
+        break;
+      }
+      break;
```
hlitz: isn't this a bug? 2nd break causes that you always only execute the loop once. Was this break in cluded to generate the last results?

src/bp/bp_targ_mech.c
hlitz: ok sounds good

src/bp/bp_targ_mech.c
hlitz: please add a comment to specifically explain that this only happens under SMC

src/bp/bp_targ_mech.c
```diff
@@ -515,21 +522,158 @@ void bp_btb_gen_recover(Bp_Data* bp_data, Recovery_Info* info) {
...
+                                                 &repl_line_addr);
+        br_slots[0] = br_slot;
+        // Invalidate the remaining slots
+        for (uns ii = 1; ii < BTB_NUM_BRSLOT; ii++)
+          br_slots[ii].valid = FALSE;
+      } else {
+        uns insert_pos = BTB_NUM_BRSLOT;
+        for (uns ii = 0; ii < BTB_NUM_BRSLOT; ii++) {
+          if (br_slots[ii].valid) {
+            // if (br_slots[ii].addr < op->inst_info->addr)
```
hlitz: remove

hlitz (changes requested):

src/bp/bp_targ_mech.c
estodi: Right, it used to be correct but I mistakenly overlooked it when making the previous commit. When slot is valid it should not break.

hlitz (dismissed):
lgtm

hlitz (approved):
LGTM

PR #498, opened 2026-04-25
L0 predictor fixes

hlitz:
stats:
- remove duplicate stats
- rename stats to use consistent naming scheme
- add new stats so that ALT and OFFPATH no longer pollute stats

fix:
- ensure L0 is predicts during warmup
- ensure L0 is predicts on off-path

main and l0 now perform the same number of predictions


src/bp/bp.c
```diff
@@ -78,10 +78,30 @@ extern void tc_do_stat(Op*, Flag);
...
+        STAT_EVENT((op)->proc_id,                                           \
+                   (op)->off_path ? MAIN_##stat##_OFFPATH : MAIN_##stat);   \
+      }                                                                     \
+    } else if (stat_mode == BP_PRED_STAT_ALT) {                             \
+      if ((pred_level) == BP_PRED_L0) {                                      \
+        STAT_EVENT((op)->proc_id, ALT_L0_##stat);                           \
+      } else {                                                              \
+        STAT_EVENT((op)->proc_id, ALT_MAIN_##stat);                         \
+      }                                                                     \
+    }                                                                       \
```
5surim: It does not need to be handled for now, but theoretically, it's possible to have four alternate BPs (five BPs in total). That said, with this patch, all the non-primary BPs will increment this ALT_L0 and ALT_MAIN stats. I had a plan to extend `STAT_EVENT` library itself to get not only proc_id, but also get bp_id to generate different stats files (e.g. `fetch.<core_id>.csv` -> `fetch.<core_id>.<bp_id>.csv` only for the frontend-affected stats) and extend infra as well.

src/bp/bp.c
```diff
@@ -193,6 +213,23 @@ void bp_sched_redirect(Bp_Recovery_Info* bp_recovery_info, Op* op, Counter cycle
...
 }
 
+/******************************************************************************/
+/* bp_stat_main_branch_resolve_latency: called when decode/exec resolves a MAIN
+ * prediction that requires frontend recovery. */
+
+void bp_stat_main_branch_resolve_latency(Op* op, Counter resolve_cycle, Flag recover_at_exec) {
+  if (op->bp_pred_info != &op->bp_pred_main)
+    return;
+  if (!(op->bp_pred_main.mispred || op->bp_pred_main.misfetch))
```
5surim: We don't want to read `mispred` and `misfetch` since we plan to remove them.

src/bp/bp.c
```diff
@@ -775,6 +827,20 @@ Addr bp_predict_op(Bp_Data* bp_data, Op* op, uns bp_id, uns br_num, Addr fetch_a
   return bp_pred_info->pred_npc;
 }
 
+Addr bp_predict_op(Bp_Data* bp_data, Op* op, uns bp_id, uns br_num, Addr fetch_addr, Bp_Pred_Level pred_level) {
+  return bp_predict_op_impl(bp_data, op, bp_id, br_num, fetch_addr, pred_level, BP_PRED_STAT_PRIMARY);
+}
+
+Addr bp_predict_op_no_stats(Bp_Data* bp_data, Op* op, uns bp_id, uns br_num, Addr fetch_addr,
```
moonseoppkim: This function doesn't have any caller, are you missing something or just for preparation?

src/decode_stage.c
```diff
@@ -272,6 +272,7 @@ void decode_stage_process_op(Op* op) {
     if (op->bp_pred_info->recover_at_decode) {
       DEBUG(dec->proc_id, "Decode schedules recovery for op_num:%llu at cycle:%llu\n", (unsigned long long)op->op_num,
             (unsigned long long)cycle_count);
+      bp_stat_main_branch_resolve_latency(op, cycle_count, FALSE);
```
moonseoppkim, line 275: bp_stat_main_branch_resolve_latency doesn't have an explicit op->off_path guard. unlike "bp_sched_recovery" function. Do you want to collect resolve latency as on & off-path mixed value?

```
void bp_sched_recovery(Bp_Recovery_Info* bp_recovery_info, Op* op, Counter cycle) {
  ASSERT(op->proc_id, bp_recovery_info->proc_id == op->proc_id);
  ASSERT(0, !op->off_path); <<
```

src/exec_stage.c
```diff
@@ -560,6 +560,7 @@ static inline void exec_stage_bp_resolve(Op* op) {
   if (op->bp_pred_info->recover_at_exec) {
     DEBUG(exec->proc_id, "Exec schedules recovery for op_num:%llu at cycle:%llu\n", (unsigned long long)op->op_num,
           (unsigned long long)op->exec_cycle);
+    bp_stat_main_branch_resolve_latency(op, op->exec_cycle, TRUE);
```
moonseoppkim, line 563: Same as above

src/bp/bp.c
hlitz: thanks for the comment. I think we can defer this to later as right now I don't see an immediate need for > 1 ALT BP

src/bp/bp.c
hlitz: done

src/decode_stage.c
hlitz: done

src/decoupled_frontend.cc
```diff
@@ -805,16 +805,24 @@ void Decoupled_FE::redirect_to_off_path(FT_PredictResult result) {
       if (per_core_dfe[proc_id][_bp_id]->get_dfe_recovery_policy() == CONTINUE_ON_PREDICTION &&
           !per_core_dfe[proc_id][_bp_id]->is_off_path()) {
         ASSERT(proc_id, !per_core_dfe[proc_id][_bp_id]->ftq_num_fts());
+        Decoupled_FE* alt_dfe = per_core_dfe[proc_id][_bp_id].get();
+        FT alt_trigger_ft(proc_id, _bp_id);
         Op alt_op = *result.op;
```
5surim, line 810: I think we should alloc_op instead of using the primary bp's result.op.
If we call alloc_op(), then bp_pred_* should be initialized.

5surim (changes requested):
Hi @hlitz ,

Thanks for the patch. Shouldn't we call alloc_op instead of reusing result.op and resetting pred fields manually?

Thanks,
Surim

5surim (approved):
LGTM

PR #502, opened 2026-04-27
[REFACTOR] Physical Issue Queue Implementation of Random Organization

yzhao321:
The original issue queue does not adapt a physical buffer, and schedules ops in an unrealistic manner. A single ready list is used for seperate reservation stations.
In this patch, a physical issue queue of random organization is implemented, to further support different scheduling methods in the future. Ops in the issue queue is unordered but age matrix pop count can be ultilized to track the relative ages of entries. Round-robin oldest first for serial and parallel picking is implemented.
This patch introduces IPC and KIPS drift.

### DIFF
--- a/src/issue_queue.cc
+++ b/src/issue_queue.cc
@@ -51,8 +51,6 @@ extern "C" {
 #include <deque>
 #include <memory>
 #include <string>
-#include <unordered_map>
-#include <unordered_set>
 #include <vector>
 
 /**************************************************************************************/
@@ -109,7 +107,6 @@ struct IssueQueueEntry {
   Op* op = nullptr;
   uns64 op_fu_type = 0;
   ISSUE_QUEUE_ENTRY_STATE state = ISSUE_QUEUE_ENTRY_STATE_EMPTY;
-  IssueQueueEntry* next_ready = nullptr;
 
   explicit IssueQueueEntry(uns16 queue_id, uns16 entry_id) : queue_id(queue_id), entry_id(entry_id) {}
   void clear();
@@ -119,7 +116,6 @@ struct IssueQueueEntry {
 void IssueQueueEntry::clear() {
   op = nullptr;
   op_fu_type = 0;
-  ASSERT(node->proc_id, next_ready == nullptr);
 }
 
 void IssueQueueEntry::fill(Op* op) {
@@ -184,14 +180,17 @@ class FunctionalUnitPicker {
   const uns32 fu_id;
   const uns64 fu_type;
 
+  std::unique_ptr<SchedulePolicy> sched_policy;
+
   // current selection for this cycle
   IssueQueueEntry* picked_entry = nullptr;
 
  public:
-  explicit FunctionalUnitPicker(uns proc_id, uns16 queue_id, uns32 fu_id, uns64 fu_type)
-      : proc_id(proc_id), queue_id(queue_id), fu_id(fu_id), fu_type(fu_type) {}
+  explicit FunctionalUnitPicker(uns proc_id, uns16 queue_id, uns32 fu_id, uns64 fu_type,
+                                std::unique_ptr<SchedulePolicy> sched_policy)
+      : proc_id(proc_id), queue_id(queue_id), fu_id(fu_id), fu_type(fu_type), sched_policy(std::move(sched_policy)) {}
 
-  void pick(IssueQueueEntry*& candidate, const SchedulePolicy& sched_policy);
+  void pick(IssueQueueEntry*& candidate);
   void grant();
   bool is_compatible(uns64 op_fu_type) const;
 
@@ -203,10 +202,8 @@ class FunctionalUnitPicker {
  * selection, it replaces the existing selection according to the
  * scheduling policy.
  */
-void FunctionalUnitPicker::pick(IssueQueueEntry*& request_entry, const SchedulePolicy& sched_policy) {
-  ASSERT(node->proc_id, request_entry != nullptr);
-  bool replace = (picked_entry == nullptr || sched_policy.compare(request_entry, picked_entry));
-  if (!replace) {
+void FunctionalUnitPicker::pick(IssueQueueEntry*& request_entry) {
+  if (!(picked_entry == nullptr || sched_policy->compare(request_entry, picked_entry))) {
     return;
   }
 
@@ -266,27 +263,21 @@ class SelectLogic {
   const uns16 queue_id;
 
   std::vector<FunctionalUnitPicker> connected_fu_pickers;
-  std::vector<std::unique_ptr<SchedulePolicy>> sched_policies;  // per-picker scheduling policies
-  std::unique_ptr<ArbitrationPolicy> arbitration_policy;        // picker traversal policy for the current cycle
+  std::unique_ptr<ArbitrationPolicy> arbitration_policy;
 
-  std::vector<size_t> picker_order;       // picker traversal order generated each cycle
-  IssueQueueEntry* ready_head = nullptr;  // head of the unsorted ready request list
-  uns64 ready_op_types = 0;               // bitmask of ready op types in this cycle
+  std::vector<size_t> picker_order;  // picker traversal order generated each cycle
+  uns64 ready_op_types = 0;          // bitmask of ready op types in this cycle
 
  public:
-  explicit SelectLogic(uns proc_id, uns16 queue_id, const std::vector<FunctionalUnitPicker>& connected_fu_pickers,
-                       std::vector<std::unique_ptr<SchedulePolicy>> sched_policies,
+  explicit SelectLogic(uns proc_id, uns16 queue_id, std::vector<FunctionalUnitPicker> connected_fu_pickers,
                        std::unique_ptr<ArbitrationPolicy> arbitration_policy)
       : proc_id(proc_id),
         queue_id(queue_id),
-        connected_fu_pickers(connected_fu_pickers),
-        sched_policies(std::move(sched_policies)),
+        connected_fu_pickers(std::move(connected_fu_pickers)),
         arbitration_policy(std::move(arbitration_policy)),
         picker_order(this->connected_fu_pickers.size()) {}
 
-  void select();
-  void request(IssueQueueEntry* entry);
-  void release(IssueQueueEntry* entry);
+  void select(std::vector<IssueQueueEntry>& entries);
 };
 
 /*
@@ -297,22 +288,22 @@ class SelectLogic {
  * Parallel selection allows each functional unit to independently select a ready op in the same cycle.
  * An arbitrator is then used to resolve conflicts when multiple functional units select the same op.
  */
-void SelectLogic::select() {
+void SelectLogic::select(std::vector<IssueQueueEntry>& entries) {
   const size_t fu_num = connected_fu_pickers.size();
   arbitration_policy->build_picker_order(picker_order);
   ready_op_types = 0;
 
-  for (IssueQueueEntry* entry = ready_head; entry != nullptr; entry = entry->next_ready) {
+  for (IssueQueueEntry& entry : entries) {
     // check if the op is ready (it may become not ready due to memory blocking or waiting for forwarding)
-    if (entry->state != ISSUE_QUEUE_ENTRY_STATE_READY || !issue_queue_check_op_ready(entry->op)) {
+    if (entry.state != ISSUE_QUEUE_ENTRY_STATE_READY || !issue_queue_check_op_ready(entry.op)) {
       continue;
     }
 
     // track the ready op types for stat collection
-    ready_op_types |= entry->op_fu_type;
+    ready_op_types |= entry.op_fu_type;
 
     // the current request propagated through the serial picker chain
-    IssueQueueEntry* request_entry = entry;
+    IssueQueueEntry* request_entry = &entry;
 
     for (size_t i = 0; i < fu_num; ++i) {
       // traverse pickers in arbitration order
@@ -322,7 +313,7 @@ void SelectLogic::select() {
       if (!fu_picker.is_compatible(request_entry->op_fu_type)) {
         continue;
       }
-      fu_picker.pick(request_entry, *sched_policies[picker_idx]);
+      fu_picker.pick(request_entry);
 
       // the request has been consumed
       if (request_entry == nullptr) {
@@ -355,33 +346,6 @@ void SelectLogic::select() {
       STAT_EVENT(node->proc_id, FU_IDLE_NO_READY_OPS_TOTAL);
       STAT_EVENT(node->proc_id, FU_0_IDLE_NO_READY_OPS + (fu_id < 32 ? fu_id : 32));
     }
-  }
-}
-
-void SelectLogic::request(IssueQueueEntry* entry) {
-  ASSERT(node->proc_id, entry->next_ready == nullptr);
-  entry->next_ready = ready_head;
-  ready_head = entry;
-}
-
-void SelectLogic::release(IssueQueueEntry* entry) {
-  IssueQueueEntry* prev = nullptr;
-  IssueQueueEntry* curr = ready_head;
-  while (curr != nullptr) {
-    if (curr != entry) {
-      prev = curr;
-      curr = curr->next_ready;
-      continue;
-    }
-
-    if (prev == nullptr) {
-      ready_head = curr->next_ready;
-    } else {
-      prev->next_ready = curr->next_ready;
-    }
-
-    curr->next_ready = nullptr;
-    return;
   }
 }
 
@@ -424,16 +388,8 @@ void RoundRobinArbitrationPolicy::build_picker_order(std::vector<size_t>& picker
 
 class IssueQueuePolicyFactory {
  public:
-  std::vector<std::unique_ptr<SchedulePolicy>> make_schedule_policies(
-      const std::vector<FunctionalUnitPicker>& fu_pickers) {
-    std::vector<std::unique_ptr<SchedulePolicy>> policies;
-    policies.reserve(fu_pickers.size());
-
-    for (size_t i = 0; i < fu_pickers.size(); i++) {
-      policies.emplace_back(std::make_unique<OldestFirstSchedulePolicy>());
-    }
-
-    return policies;
+  std::unique_ptr<SchedulePolicy> make_schedule_policy(uns16 queue_id, uns32 fu_id, uns64 fu_type) {
+    return std::make_unique<OldestFirstSchedulePolicy>();
   }
 
   std::unique_ptr<ArbitrationPolicy> make_arbitration_policy(size_t fu_num) {
@@ -457,8 +413,7 @@ class IssueQueue {
   std::unique_ptr<SelectLogic> select_logic;
 
  public:
-  explicit IssueQueue(uns proc_id, uns16 queue_id, uns16 size,
-                      const std::vector<FunctionalUnitPicker>& connected_fu_pickers);
+  explicit IssueQueue(uns proc_id, uns16 queue_id, uns16 size, std::vector<FunctionalUnitPicker> connected_fu_pickers);
   uns16 allocate_entry(Op* op);
   void free_entry(uns16 entry_id);
   size_t available_entries() const { return free_list.size(); }
@@ -472,8 +427,7 @@ class IssueQueue {
   void schedule();
 };
 
-IssueQueue::IssueQueue(uns proc_id, uns16 queue_id, uns16 size,
-                       const std::vector<FunctionalUnitPicker>& connected_fu_pickers)
+IssueQueue::IssueQueue(uns proc_id, uns16 queue_id, uns16 size, std::vector<FunctionalUnitPicker> connected_fu_pickers)
     : proc_id(proc_id), queue_id(queue_id) {
   // initialize entries and free list
   entries.reserve(size);
@@ -485,8 +439,7 @@ IssueQueue::IssueQueue(uns proc_id, uns16 queue_id, uns16 size,
 
   // initialize select logic from the scheduling and arbitration policies.
   IssueQueuePolicyFactory factory;
-  select_logic = std::make_unique<SelectLogic>(proc_id, queue_id, connected_fu_pickers,
-                                               factory.make_schedule_policies(connected_fu_pickers),
+  select_logic = std::make_unique<SelectLogic>(proc_id, queue_id, std::move(connected_fu_pickers),
                                                factory.make_arbitration_policy(connected_fu_pickers.size()));
 }
 
@@ -524,7 +477,6 @@ void IssueQueue::wakeup(uns16 entry_id) {
 
   op->in_rdy_list = TRUE;
   entry.state = ISSUE_QUEUE_ENTRY_STATE_READY;
-  select_logic->request(&entry);
 }
 
 void IssueQueue::issued(uns16 entry_id) {
@@ -537,7 +489,6 @@ void IssueQueue::issued(uns16 entry_id) {
 
   STAT_EVENT(node->proc_id, OP_ISSUED);
   op->in_rdy_list = FALSE;
-  select_logic->release(&entry);
   free_entry(entry.entry_id);
 }
 
@@ -561,13 +512,12 @@ void IssueQueue::recover() {
     }
 
     op->in_rdy_list = FALSE;
-    select_logic->release(&entry);
     free_entry(entry.entry_id);
   }
 }
 
 void IssueQueue::schedule() {
-  select_logic->select();
+  select_logic->select(entries);
 }
 
 /**************************************************************************************/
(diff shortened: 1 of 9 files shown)


src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+
+#define DEBUG(proc_id, args...) _DEBUG(proc_id, DEBUG_NODE_STAGE, ##args)
+
+/**************************************************************************************/
+/* Constexpr */
+
+enum ISSUE_QUEUE_ENTRY_STATE {
+  ISSUE_QUEUE_ENTRY_STATE_EMPTY,
+  ISSUE_QUEUE_ENTRY_STATE_ALLOC,
+  ISSUE_QUEUE_ENTRY_STATE_READY,
```
hlitz, line 68: do we need and ISSUED state? You said that sometimes stuff gets issued but we cannot immediately remove it from IQ. We do not want to issue an op twices that is still READY

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+/* Macros */
+
+#define DEBUG(proc_id, args...) _DEBUG(proc_id, DEBUG_NODE_STAGE, ##args)
+
+/**************************************************************************************/
+/* Constexpr */
+
+enum ISSUE_QUEUE_ENTRY_STATE {
+  ISSUE_QUEUE_ENTRY_STATE_EMPTY,
+  ISSUE_QUEUE_ENTRY_STATE_ALLOC,
```
hlitz: rename to ALLOCED.

hlitz: ALLOC is not clear whether it is ready for alloc or already allocated

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+                                    const std::unordered_set<Counter>& picked_mask_curr) {
+  while (ready_iter != ready_list.end()) {
+    Op* op = ready_iter->second;
+    ++ready_iter;
+
+    /*
+     * Skip ops that were already issued in previous rounds.
+     * These are globally visible to all pickers this round, so we can safely advance.
+     */
+    if (picked_mask_prev.count(op->op_num)) {
```
hlitz: I think it would be cleaner to check for sth like if(op->state == ISUED)

because we may want to cancel ops after 2 or 3 cycles. THen we would have to have mulitple prev masks and shift them around.

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+      continue;
+    }
+
+    /*
+     * Conflict with another picker in the current round.
+     * This picker does not observe current-round selections, so arbitration may cancel it.
+     * Stop here and defer to the next round to avoid inconsistent ordering.
+     */
+    if (picked_mask_curr.count(op->op_num)) {
+      break;
```
hlitz: we could use e.g. if (op->state == PICKED) instead but we would need a mechansim that at the end of a round turns all PICKED into ISSUED.

I thought we would only break in parallel mode but continue in serial mode. Or is the serial not yet implemented?

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+
+ public:
+  OldestFirstSelectLogic(uns queue_id, const std::vector<FunctionalUnitPicker>& connected_fu_pickers)
+      : queue_id(queue_id), fu_num(connected_fu_pickers.size()), connected_fu_pickers(connected_fu_pickers) {}
+  void request(Op* op) override;
+  void release(Op* op) override;
+  void select() override;
+};
+
+void OldestFirstSelectLogic::request(Op* op) {
```
hlitz: can we move this into parent class selectLogic? It does not seem to be oldestfirst specific

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+  for (size_t i = 0; i < fu_num; ++i) {
+    FunctionalUnitPicker& fu_picker = connected_fu_pickers[i];
+    if (fu_picker.is_compatible(op->inst_info)) {
+      fu_picker.add_ready_op(op->op_num, op);
+    }
+  }
+}
+
+void OldestFirstSelectLogic::release(Op* op) {
+  op->in_rdy_list = FALSE;
```
hlitz: same. move into selectLogic?

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+  for (size_t i = 0; i < fu_num; ++i) {
+    FunctionalUnitPicker& fu_picker = connected_fu_pickers[i];
+    fu_picker.remove_ready_op(op->op_num);
+  }
+}
+
+/*
+ * Serial selection is described in "Quantifying the complexity of superscalar processors", 1996.
```
hlitz, line 288: can we make serial/parallel independent from the pick-order logic? I.e. we have pick order oldest,random,crit,type and all four can be either parallel or serial? 

(not all may make sense but we could still keep it generic)

so we have a pick_order policy and a pick_implementation policy, and a pick_fu_order policy.

pick_order_policy (oldest,crit,random,type): Determines how ops are inserted in ready_list
pick_implementation_policy (serial,parallel): Switches between break/continue in op->state==PICKED
pick_fu_order_policy (round robin,random): determines  picker_order vector

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+
+    for (size_t picker_idx : picker_order) {
+      if (!(active_mask & (1ULL << picker_idx))) {
+        continue;
+      }
+
+      FunctionalUnitPicker& fu_picker = connected_fu_pickers[picker_idx];
+      Op* op = fu_picker.pick_next(picked_mask_prev, picked_mask_curr);
+      if (op == nullptr) {
+        if (!fu_picker.has_ready_op()) {
```
hlitz: move up this check? ie. do not call pick_next if !has_ready_op()

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+  ASSERT(proc_id, entry.state == ISSUE_QUEUE_ENTRY_STATE_ALLOC);
+
+  Op* op = entry.op;
+  ASSERT(proc_id, op != nullptr);
+
+  entry.state = ISSUE_QUEUE_ENTRY_STATE_READY;
+  select_logic->request(op);
+}
+
+void IssueQueue::grant() {
```
hlitz: can we have the ops themselves call grant() ?  I.e. wheneveer an op changes its op->state to issued in the exec stage it calls into the scheduler to notify successful acceptance by the FU?

Also I m not sure grant is the right name here. In the context of a scheduler, one would think that the ready ops request and the scheduler grants. But here it is that the FU is requested and the FU grants, this is a bit ambigous. so we could just call it IQ->issued(op)

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+      continue;
+    }
+
+    STAT_EVENT(node->proc_id, OP_ISSUED);
+    select_logic->release(op);
+    free_entry(entry.entry_id);
+  }
+}
+
+void IssueQueue::recover() {
```
hlitz, line 520: could we provide the flush_id and then check:
if (op->off_path) op->uid > flush_id
else op->uild <= flush_id

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+ *  - Allocate it to its designated reservation station.
+ *  - If all source operands are ready, insert it into the ready list.
+ */
+void IssueQueues::dispatch() {
+  Op* op = NULL;
+  uns32 num_fill_rs = 0;
+
+  for (op = node->next_op_into_rs; op; op = op->next_node) {
+    ASSERT(proc_id, op->queue_id == MAX_UNS16 && op->queue_entry_id == MAX_UNS16);
+    uns16 queue_id = find_emptiest_queue(op);
```
hlitz, line 640: should also depend on dispatch_policy but we can do that later

src/issue_queue.cc
```diff
@@ -0,0 +1,697 @@
...
+    if (queue_id == MAX_UNS16) {
+      break;
```
hlitz, line 642: do we want to add a stat to count stalls? or is this already done in the caller of dispatch()

hlitz (changes requested):
looks very clean but some comments

src/issue_queue.cc
yzhao321: renamed.

src/issue_queue.cc
yzhao321: Updated as cancel_mask + pick_state

src/issue_queue.cc
yzhao321: use exec.issued_func to release PICKED. and use exec.reject_func to change PICKED back to READY

src/issue_queue.cc
yzhao321: Request is diff for each specific select_logic.
Age: ready_list.insert(op->op_num)
Random: ready_list.insert(op->queue_entry_id)
Crit: ready_list.insert(op->weight)

src/issue_queue.cc
yzhao321: It may be common. But I keep it parallel with req

src/issue_queue.cc
yzhao321: Updated. Each select_logic needs to specify its picker_order func and request (ready_list insertion) func

src/issue_queue.cc
yzhao321: It is to handle the case when the picker do not have ready ops but need to break the loop. Now it is changed to let pick_next return the status.

src/issue_queue.cc
yzhao321: Call back mechanism is introduced. Exec will call issued or reject in exec_stage

src/issue_queue.cc
yzhao321: Added. Just a question, will we introduce flush part of off-path ops in the future?

src/issue_queue.cc
yzhao321: There is not counter in the caller. We can add the stat in the patch that let dispatch depends on dispatch_policy

src/issue_queue.cc
```diff
@@ -0,0 +1,755 @@
...
+  void remove_ready_entry(uns64 key) { ready_list.erase(key); }
+  void init_ready_iter() { ready_iter = ready_list.begin(); }
+};
+
+IssueQueueEntry* FunctionalUnitPicker::pick_next(const std::unordered_set<Counter>& picked_mask) {
+  while (ready_iter != ready_list.end()) {
+    IssueQueueEntry* entry = ready_iter->second;
+    Op* op = entry->op;
+    ++ready_iter;
+
```
hlitz: don't we have to check for is_compatible() at the beginning now?  as we share the IQ between FUs and some may not be able to handle both.

src/issue_queue.cc
```diff
@@ -0,0 +1,755 @@
...
+    if (entry->state == ISSUE_QUEUE_ENTRY_STATE_PICKED) {
+      /*
+       * Conflict with another picker in the same round (parallel selection).
+       * This picker is unaware of current-round selections and loses arbitration.
+       * The selection is cancelled and deferred to the next round.
+       */
+      if (picked_mask.count(entry->entry_id)) {
```
hlitz: I suggest we get rid of the picked_mask and check for PARALLEL/SERIAL. 

Because now the pick mask is parallel only, so it doesn't provide a benefit and it may be confusing if we do not set the pick_mask for SERIAL

src/issue_queue.cc
```diff
@@ -0,0 +1,759 @@
...
+  IssueQueueEntry* pick(IssueQueueEntry* ready_head, const std::unordered_set<uns16>& picked_mask,
+                        const SchedulePolicy& sched_policy);
+  void grant(IssueQueueEntry* entry);
+  bool is_compatible(Inst_Info* inst_info) const;
+};
+
+IssueQueueEntry* FunctionalUnitPicker::pick(IssueQueueEntry* ready_head, const std::unordered_set<uns16>& picked_mask,
+                                            const SchedulePolicy& sched_policy) {
+  IssueQueueEntry* ret = nullptr;
+  for (IssueQueueEntry* entry = ready_head; entry != nullptr; entry = entry->next_ready) {
```
hlitz: are you still assuming the issuequeueEntries to be ordered by readyness? or what is the meaning of a ready head, if the queue contains both ready and not ready entries in random order?

src/issue_queue.cc
```diff
@@ -0,0 +1,759 @@
...
+    if (entry->state == ISSUE_QUEUE_ENTRY_STATE_PICKED) {
```
hlitz: I thought we said that we would break in the parallel case here?

src/issue_queue.cc
```diff
@@ -0,0 +1,759 @@
...
+    FunctionalUnitPicker& fu_picker = connected_fu_pickers[arbitration_policy->picker_at(i)];
+    IssueQueueEntry* entry = fu_picker.pick(ready_head, picked_mask, *sched_policy);
+
+    // if the picking is cancelled due to a conflict in the same round or there is no compatible entry
+    if (entry == nullptr) {
+      continue;
+    }
+
+    // mark the current pick as the winner to cancel competing picks for the same op in this round
+    if (ISSUE_QUEUE_PARALLEL_PICK) {
```
hlitz: I think you don't need a pick mask if you just return instead of continuing here

 if (entry->state == ISSUE_QUEUE_ENTRY_STATE_PICKED) {
       if (ISSUE_QUEUE_PARALLEL_PICK) return null; //you have to reorder and check for compatibility first
       continue;

src/issue_queue.cc
hlitz: not sure in memtrace there is no way to know if a off-path branch resolves "correctly". For exec it can be done in principle but I think it is tricky as it requires multiple checkpoints etc.

src/issue_queue.cc
hlitz: ok the ready-list implementation is ok but I still don't get the parallel implementation. In parallel you need to allow muitple FUs to pick the same op so here you can only continue if you are in serial mode. In parallel mode you ignore what has already been picked. Parallel needs to look at all ops (including picked) then makes it pick decsision, then if there is a duplicate it gets cancelled.

src/issue_queue.cc
hlitz: I think you want to always insert in SERIAL  and PARALLEL

hlitz (changes requested):

src/issue_queue.cc
```diff
@@ -0,0 +1,840 @@
...
+ * and an ArbitrationPolicy, which determines the priority of pickers during selection.
+ */
+
+class SelectLogic {
+ protected:
+  const uns proc_id;
+  const uns16 queue_id;
+
+  std::vector<FunctionalUnitPicker> connected_fu_pickers;
+  std::vector<std::unique_ptr<SchedulePolicy>> sched_policies;  // per-picker scheduling policies
```
hlitz: if the policy is part a property of each piccker why is it not a member variable of FunctionalUnitPicker?

src/issue_queue.cc
```diff
@@ -0,0 +1,840 @@
...
+      // the request has been consumed
+      if (request_entry == nullptr) {
+        break;
+      }
+    }
+
+    // the entry is not picked or displaced
+    if (request_entry != nullptr) {
+      STAT_EVENT(node->proc_id, RS_OP_READY_NOT_ISSUED_TOTAL);
+      STAT_EVENT(node->proc_id, RS_0_OP_READY_NOT_ISSUED + (queue_id < 8 ? queue_id : 8));
```
hlitz, line 331: what is the magic number 8 here? do we only support stats for up to 8 queues?

use a constant (e.g. MAX_QUEUES) and introduce an init assertion that the number of queues is equal or less to MAX_QUEUES. We should probably increase MAX_QUEUES to e.g. 32. A ten wide machine could already use 10 queues, so 8 does not seem enough

src/issue_queue.cc
```diff
@@ -0,0 +1,840 @@
...
+    uns32 fu_id = fu_picker.get_fu_id();
+    if (node->sd.ops[fu_id] != NULL)
+      continue;
+
+    Func_Unit* fu = &exec->fus[fu_id];
+    if (fu->avail_cycle > cycle_count || fu->held_by_mem)
+      continue;
+
+    if ((ready_op_types & fu->type) == 0) {
+      STAT_EVENT(node->proc_id, FU_IDLE_NO_READY_OPS_TOTAL);
```
hlitz, line 350: can we have 3 stats that sum up to 100% ?

ie.
RDY_AND_NOT_PICKED
RDY_AND_PICKED
NOT_RDY

src/issue_queue.cc
```diff
@@ -0,0 +1,840 @@
...
+void SelectLogic::request(IssueQueueEntry* entry) {
+  ASSERT(node->proc_id, entry->next_ready == nullptr);
+  entry->next_ready = ready_head;
+  ready_head = entry;
+}
+
+void SelectLogic::release(IssueQueueEntry* entry) {
+  IssueQueueEntry* prev = nullptr;
+  IssueQueueEntry* curr = ready_head;
+  while (curr != nullptr) {
```
hlitz: can we use std::list instead of rolling our own implementation? The ready-list should be independent of the node list right?

src/issue_queue.cc
```diff
@@ -0,0 +1,840 @@
...
+
+  // initialize connection map from FUs to issue queues
+  std::vector<std::string> connection_tokens = split_tokens(RS_CONNECTIONS);
+  ASSERT(proc_id, connection_tokens.size() == NUM_RS);
+  fu_map.assign(NUM_FUS, MAX_UNS16);
+  for (size_t i = 0; i < NUM_RS; ++i) {
+    for (uns64 mask = parse_mask(connection_tokens[i]); mask; mask &= (mask - 1)) {
+      int idx = __builtin_ctzll(mask);
+      ASSERT(proc_id, idx < (int)NUM_FUS);
+      fu_map[idx] = i;
```
hlitz: I am not sure if I get this. First, it seems possible that we assign the same fu_map[idx] twice but this should not happen. Maybe we want to assert ASSERT(proc_id, fu_map[idx] == 0); (or INVALID).

Second, is RS_CONECTIONS used anywhere else? Wouldn't it be better to have FU_CONNECTIONS instead where each FU would list exactly one RS (instead of a bitmap)?

hlitz: third, would be good to assert that every fu has been mapped (exactly once)

hlitz (changes requested):

src/issue_queue.cc
yzhao321: The stats and numbers are from this PR: https://github.com/litz-lab/scarab_ll/pull/373
I would prefer to update these stats in another PR

src/issue_queue.cc
yzhao321: FU_IDLE_NO_READY_OPS_TOTAL this stat is calculated based on FU not op.
To sum up to 100%, need to caculate on the op loop side.
I prefer to update the stats in another PR.

src/issue_queue.cc
yzhao321: related ASSERTION is added.
RS_CONECTIONS is used in exec_port.c
We updated it in another PR.

src/issue_queue.cc
yzhao321: Moved. The reason is that I thought the policy should belong to select logic logically.

src/issue_queue.cc
yzhao321: I directly removed the ready_list and select func will scan the whole queue.
I don't know if std::list will result in perf drift as it will alloc/free.
ready list is independent of node list.

yzhao321: | Workload | baseline (SIM_HOST_WALL_SECONDS) | parallel_queue (SIM_HOST_WALL_SECONDS) | serial_queue (SIM_HOST_WALL_SECONDS) | parallel_queue speedup vs baseline (%) | serial_queue speedup vs baseline (%) |
| --- | --- | --- | --- | --- | --- |
| spec2017/rate_int_v2/deepsjeng_r | 821.3309 | 1018.6846 | 844.4234 | 24.03% | 2.81% |
| spec2017/rate_int_v2/exchange2_r | 560.3008 | 811.1045 | 645.4435 | 44.76% | 15.2% |
| spec2017/rate_int_v2/gcc_r | 1023.6187 | 1363.7754 | 1190.6982 | 33.23% | 16.32% |
| spec2017/rate_int_v2/gcc_r_2 | 1259.5653 | 1337.209 | 1207.5138 | 6.16% | -4.13% |
| spec2017/rate_int_v2/gcc_r_3 | 900.3904 | 1132.3516 | 1010.0721 | 25.76% | 12.18% |
| spec2017/rate_int_v2/gcc_r_4 | 854.3924 | 1115.4774 | 1026.1026 | 30.56% | 20.1% |
| spec2017/rate_int_v2/gcc_r_5 | 795.9238 | 1169.3592 | 1072.3925 | 46.92% | 34.74% |
| spec2017/rate_int_v2/leela_r | 898.799 | 1188.2643 | 975.8022 | 32.21% | 8.57% |
| spec2017/rate_int_v2/mcf_r | 1095.746 | 1462.7866 | 1314.5214 | 33.5% | 19.97% |
| spec2017/rate_int_v2/omnetpp_r | 1100.7578 | 1141.3566 | 1210.496 | 3.69% | 9.97% |
| spec2017/rate_int_v2/perlbench_r | 828.1971 | 1082.0821 | 943.0243 | 30.66% | 13.86% |
| spec2017/rate_int_v2/perlbench_r_2 | 808.1187 | 1073.3573 | 929.4358 | 32.82% | 15.01% |
| spec2017/rate_int_v2/perlbench_r_3 | 738.8204 | 973.9845 | 904.9564 | 31.83% | 22.49% |
| spec2017/rate_int_v2/x264_r | 380.7947 | 592.1094 | 488.7229 | 55.49% | 28.34% |
| spec2017/rate_int_v2/x264_r_2 | 388.9108 | 575.2437 | 479.6283 | 47.91% | 23.33% |
| spec2017/rate_int_v2/x264_r_3 | 433.4535 | 588.0507 | 487.1421 | 35.67% | 12.39% |
| spec2017/rate_int_v2/xalancbmk_r | 787.6825 | 840.0799 | 788.5115 | 6.65% | 0.11% |
| spec2017/rate_int_v2/xz_r | 759.4309 | 1060.0921 | 952.6787 | 39.59% | 25.45% |
| spec2017/rate_int_v2/xz_r_2 | 433.0658 | 584.8111 | 537.4539 | 35.04% | 24.1% |
| spec2017/rate_int_v2/xz_r_3 | 659.7616 | 437.6134 | 352.5612 | -33.67% | -46.56% |
| Avg | 776.4531 | 977.3897 | 868.079 | 25.88% | 11.8% |

Although queue-based implementation is not shown in perf, it introduce decrement.

yzhao321: std::list introduce perf downgrad

| Workload | baseline (SIM_HOST_WALL_SECONDS) | parallel_std_list (SIM_HOST_WALL_SECONDS) | serial_std_list (SIM_HOST_WALL_SECONDS) | parallel_std_list speedup vs baseline (%) | serial_std_list speedup vs baseline (%) |
| --- | --- | --- | --- | --- | --- |
| spec2017/rate_int_v2/deepsjeng_r | 821.3309 | 1014.1977 | 861.8727 | 23.48% | 4.94% |
| spec2017/rate_int_v2/exchange2_r | 560.3008 | 822.1572 | 653.0089 | 46.73% | 16.55% |
| spec2017/rate_int_v2/gcc_r | 1023.6187 | 1364.6322 | 1217.2284 | 33.31% | 18.91% |
| spec2017/rate_int_v2/gcc_r_2 | 1259.5653 | 1356.5168 | 1201.5574 | 7.7% | -4.61% |
| spec2017/rate_int_v2/gcc_r_3 | 900.3904 | 1131.2944 | 1012.0277 | 25.64% | 12.4% |
| spec2017/rate_int_v2/gcc_r_4 | 854.3924 | 1120.1468 | 1042.1639 | 31.1% | 21.98% |
| spec2017/rate_int_v2/gcc_r_5 | 795.9238 | 1169.2801 | 1053.8704 | 46.91% | 32.41% |
| spec2017/rate_int_v2/leela_r | 898.799 | 1197.4646 | 980.9385 | 33.23% | 9.14% |
| spec2017/rate_int_v2/mcf_r | 1095.746 | 1457.3623 | 1344.3712 | 33% | 22.69% |
| spec2017/rate_int_v2/omnetpp_r | 1100.7578 | 1156.1379 | 1230.0661 | 5.03% | 11.75% |
| spec2017/rate_int_v2/perlbench_r | 828.1971 | 1109.3385 | 912.2488 | 33.95% | 10.15% |
| spec2017/rate_int_v2/perlbench_r_2 | 808.1187 | 1148.0865 | 954.3745 | 42.07% | 18.1% |
| spec2017/rate_int_v2/perlbench_r_3 | 738.8204 | 1012.3511 | 896.5342 | 37.02% | 21.35% |
| spec2017/rate_int_v2/x264_r | 380.7947 | 588.2139 | 494.6299 | 54.47% | 29.89% |
| spec2017/rate_int_v2/x264_r_2 | 388.9108 | 579.1245 | 479.6782 | 48.91% | 23.34% |
| spec2017/rate_int_v2/x264_r_3 | 433.4535 | 588.8658 | 493.6414 | 35.85% | 13.89% |
| spec2017/rate_int_v2/xalancbmk_r | 787.6825 | 847.2131 | 774.9137 | 7.56% | -1.62% |
| spec2017/rate_int_v2/xz_r | 759.4309 | 1074.4386 | 947.8238 | 41.48% | 24.81% |
| spec2017/rate_int_v2/xz_r_2 | 433.0658 | 580.2268 | 539.3619 | 33.98% | 24.55% |
| spec2017/rate_int_v2/xz_r_3 | 659.7616 | 434.1824 | 352.2891 | -34.19% | -46.6% |
| Avg | 776.4531 | 987.5616 | 872.13 | 27.19% | 12.32% |

src/issue_queue.cc
hlitz: ok

src/issue_queue.cc
hlitz: ok

hlitz (approved):

PR #503, opened 2026-04-28
removed callsys check in stall and resolve

yql5510718:
Make IGNORE_BAR_FETCH the single switch for syscall fetch barriers
Drop `IS_CALLSYS` from the matched-pair sites so `BAR_FETCH` is the only
no pref_drift


hlitz (approved):
LGTM

hlitz (approved):

PR #504, opened 2026-04-30
Adjust recovery2

hlitz:
rebased the L0-BTB latency fix


src/bp/bp_targ_mech.c
5surim, line 347: pc_plus_offset == oracle.target no longer treated as a BTB hit because btb_pred_latency = MAX_UNS.
I think we should also assign `btb_pred_latency = BTB_L0_LATENCY` here as well.

src/ft.cc
```diff
@@ -352,21 +352,12 @@ FT_PredictResult FT::predict_ft() {
...
-            // recovery is bounded by the direction prediction, not the target.
-            const Flag l0_btb_failed_target =
-                !op->btb_pred_info->btb_l0_hit || op->btb_pred_info->btb_l0_target != correct_target;
-            if (BTB_L1_PRESENT && l0_btb_failed_target && op->btb_pred_info->btb_l1_hit &&
-                op->btb_pred_info->btb_l1_target == correct_target) {
-              recovery_cycle = fetch_cycle + BTB_L1_LATENCY;
-              STAT_EVENT(proc_id, BTB_L1_HIT_SCHED_EARLY_RECOVERY);
+            if (op->btb_pred_info->btb_pred_latency != MAX_UNS) {
+              const uns btb_after_l0 = btb_latency > BP_L0_LATENCY ? btb_latency - BP_L0_LATENCY : 0;
+              recovery_cycle = fetch_cycle + MAX2(BP_MAIN_LATENCY, btb_after_l0);
```
5surim: Could this just be MAX2(BP_MAIN_LATENCY, btb_latency)?

The btb_latency - BP_L0_LATENCY form is hard to follow, and I think it undercounts in the slow-BTB case:
With btb_latency = 5, BP_MAIN_LATENCY = 3, BP_L0_LATENCY = 1, the current expression gives fetch + 4, but the BTB target isn't really available until fetch + 5.

src/bp/bp.c
```diff
@@ -390,10 +390,15 @@ static Addr bp_predict_op_impl(Bp_Data* bp_data, Op* op, uns bp_id, uns br_num,
...
-  // stored in op->btb_pred_info.  bp_predict_op() is a pure reader of
-  // btb_pred_info; it does not write to it.
-  pred_target = op->btb_pred_info->pred_target;
-  Flag btb_hit = !op->btb_pred_info->btb_miss;
+  // stored in op->btb_pred_info.  The BTB target is usable by this predictor
+  // level only if the BTB level that supplied it is no slower than the BP.
+  const uns bp_latency = pred_level == BP_PRED_L0 ? BP_L0_LATENCY : BP_MAIN_LATENCY;
+  const Flag btb_hit =
+      op->btb_pred_info->btb_pred_latency != MAX_UNS && op->btb_pred_info->btb_pred_latency <= bp_latency;
+  if (btb_hit)
```
5surim, line 402: Could we centralize this in `bp_predict_btb` and derive `op->btb_pred_info->btb_miss` from `btb_pred_latency` at the end of prediction? Right now, btb_miss and btb_pred_latency are set independently and can drift out of sync (the same kind of redundancy we already had between mispred/misfetch and recover_at_decode/recover_at_exec).

With a single source of truth (btb_miss = (btb_pred_latency == MAX_UNS)), btb_pred_info cleanly captures the per-op fact "did any BTB level hit."
`bp_predict_op_impl` then makes the per-BP-level decision by combining both field `!btb_miss && btb_pred_latency <= bp_latency`.

src/confidence/conf.cc
```diff
@@ -10,12 +10,14 @@
 void ConfMechStatBase::per_cycle_update(Conf_Off_Path_Reason reason) {
   if (conf_off_path_reason == REASON_CONF_NOT_IDENTIFIED && reason != REASON_CONF_NOT_IDENTIFIED)
     conf_off_path_reason = reason;
+  const Off_Path_Reason stat_off_path_reason =
+      (off_path_reason == REASON_LATE_BTB_HIT) ? REASON_NOT_IDENTIFIED : off_path_reason;
```
5surim: Could we introduce a dedicated stat for REASON_LATE_BTB_HIT instead of collapsing it into NOT_IDENTIFIED? NOT_IDENTIFIED is supposed to mean "we couldn't classify the off-path reason," so folding a known reason into it makes the bucket misleading for debugging. Adding a DFE_OFF_CONF_ON_LATE_BTB_HIT_* family keeps the classification correct.

5surim (changes requested):
Hi @hlitz ,

I think there is a missing latency update.
Could you have a look at the comments?

Thanks,
Surim

src/ft.cc
hlitz: I think its ok because I was thinking we define all latencies relative to BP_L0_LATENCY.

For instance, if we have BP_L0==hit and BTB_L0==hit we have 0 recovery cycles (no recovery at all) although L0 latencies are 1. Now if we have latencies of 3 (L1_BTB or BP_MAIN) then the recovery latency should be 2 more than in the hit case because BTB and BP have 2 cycles longer latency.

src/bp/bp_targ_mech.c
hlitz: done

src/confidence/conf.cc
hlitz: done

src/bp/bp.c
hlitz: yes that makes sense. I pushed another update

5surim: Is btb_pred_info->btb_miss not used/set anywhere now? Then, can we just remove it?

src/ft.cc
5surim: Then shouldn't it be `MAX2(BP_MAIN_LATENCY - BP_L0_LATENCY, btb_after_l0)` instead of `MAX2(BP_MAIN_LATENCY, btb_after_l0)`?

5surim: btw this is already in `if (l0_wrong && !main_wrong) {` so BP_L0 hit is not the case here.

hlitz: yes. please have another look.

5surim (dismissed):
LGTM

5surim (approved):
LGTM.

Do you have DC numbers by any chance?
They might be more IPC drop in DC.

PR #513, opened 2026-05-14
issue_queue: silence Clang -Wunused-private-field for queue_id

5surim:
- Surfaced by running `./sci --lint-scarab` locally on main, which reported two clang-diagnostic-unused-private-field errors in src/issue_queue.cc (lines 183 and 431).
 - Adds `(void)this->queue_id;` to the `FunctionalUnitPicker` and `IssueQueue` constructor bodies so `clang-diagnostic-unused-private-field` (Clang-only; surfaced via `--lint-scarab` / clang-tidy) stops firing on the two private `queue_id` members.
  - The members themselves are kept in place — they are reserved for upcoming use, so removing them would force re-adding the field, the constructor parameter on `FunctionalUnitPicker`, and the call site in `IssueQueues::IssueQueues` in a follow-up PR.
  - `[[maybe_unused]]` was considered first (the textbook C++17 spelling), but GCC silently ignores the attribute on non-static data members, and our build enables `-Werror=attributes`, which turns the resulting "attribute ignored" warning into a hard build error. The
  `(void)` cast achieves the same effect portably.

Not sure why the CI workflow (--lint-scarab) did not catch this.


yzhao321 (approved):
LGTM

PR #515, opened 2026-05-18
[FEATURE] Pick Bucket for Parallel Picking

yzhao321:


src/issue_queue.cc
```diff
@@ -142,7 +144,8 @@ class SchedulePolicy {
   virtual ~SchedulePolicy() = default;
 
   // Return if the lhs entry is the preference
-  virtual bool compare(const IssueQueueEntry* lhs, const IssueQueueEntry* rhs) const = 0;
+  virtual bool compare(const IssueQueueEntry* lhs, const IssueQueueEntry* rhs,
```
hlitz, line 147: it seems inelegant having to pass fu into the compare func. Maybe store "parent_fu" as part of schedpolicy and then the compare can access it?

src/issue_queue.cc
```diff
@@ -209,8 +208,9 @@ class FunctionalUnitPicker {
  * selection, it replaces the existing selection according to the
  * scheduling policy.
  */
-void FunctionalUnitPicker::pick(IssueQueueEntry*& request_entry) {
-  if (!(picked_entry == nullptr || sched_policy->compare(request_entry, picked_entry))) {
+void FunctionalUnitPicker::pick(IssueQueueEntry*& request_entry, const SchedulePolicy& sched_policy) {
```
hlitz, line 211: not needed if we fix my first comment

hlitz (changes requested):

PR #516, opened 2026-05-23
[FEATURE] Issue Early Binding During Dispatching

yzhao321:
This patch introduces issue early binding for parallel picking.
1. Since there is no parallel picking conflict, ENTRY_PICKED status and corresponding 'reject' func is removed.
2. Each select_logic have only one sched_policy.
3. Rename the ArbitrationPolicy as TraversalPolicy as it is only used for serial picking. Arbitration for parallel picking  depends on BindPolicy.

Early Binding is defautly enabled. Perf drift is introduced.

### DIFF
--- a/src/issue_queue.cc
+++ b/src/issue_queue.cc
@@ -222,11 +222,6 @@ void FunctionalUnitPicker::pick(IssueQueueEntry*& request_entry, const ScheduleP
   IssueQueueEntry* displaced_entry = picked_entry;
   picked_entry = request_entry;
 
-  if (ISSUE_QUEUE_FU_EARLY_BIND) {
-    request_entry = nullptr;
-    return;
-  }
-
   // forward the displaced request to later pickers
   request_entry = displaced_entry;
 }
@@ -331,7 +326,7 @@ void SelectLogic::select() {
       }
 
       // if early binding is enabled, the request can only be picked by the bound picker
-      if (ISSUE_QUEUE_FU_EARLY_BIND && picker_idx != entry->bound_fu_id) {
+      if (entry->bound_fu_id != MAX_UNS && entry->bound_fu_id != picker_idx) {
         continue;
       }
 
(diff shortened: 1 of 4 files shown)


src/issue_queue.cc
```diff
@@ -388,37 +411,86 @@ bool OldestFirstSchedulePolicy::compare(const IssueQueueEntry* lhs, const IssueQ
...
+      best_bound_ops = bound_op_counts[i];
+    }
+  }
+
+  ASSERT(node->proc_id, best_picker_idx != MAX_UNS);
+  entry->bound_port = best_picker_idx;
+  bound_op_counts[best_picker_idx] += 1;
+}
+
+void LeastBoundBindPolicy::unbind(IssueQueueEntry* entry) {
```
hlitz: do we really need to unbind? can we have an issued() API that gets called after pick and that does various things inlcuding "unbind"

src/issue_queue.cc
```diff
@@ -388,37 +411,86 @@ bool OldestFirstSchedulePolicy::compare(const IssueQueueEntry* lhs, const IssueQ
...
+  size_t best_picker_idx = MAX_UNS;
+  size_t best_bound_ops = 0;
+
+  for (size_t i = 0; i < connected_ports.size(); ++i) {
+    const IssuePort& issue_port = connected_ports[i];
+    if (!issue_port.is_compatible(entry->op_fu_type)) {
+      continue;
+    }
+
+    if (best_picker_idx == MAX_UNS || bound_op_counts[i] < best_bound_ops) {
```
hlitz: i think "best" is not clear. If i understand the policy it checks for emptiest compatible? If yes call it emptiest and not best.

src/issue_queue.cc
```diff
@@ -146,27 +149,32 @@ class SchedulePolicy {
...
+  virtual ~BindPolicy() = default;
+  virtual void bind(const std::vector<IssuePort>& connected_ports, IssueQueueEntry* entry) = 0;
+  virtual void unbind(IssueQueueEntry* entry) = 0;
+};
+
 /**************************************************************************************/
 
 /*
- * FunctionalUnitPicker selects a single op for a functional unit during the current
+ * IssuePort selects a single op for a functional unit during the current
```
hlitz: why do we rename FU to port? Is this just a refactor or is port != FU ?
If its truely the samec thing, I'd suggest to stay with FU. I don't see the reason for introducing a new term. If its not the same than some description defining the difference is needed.

hlitz (dismissed):
LGTM

hlitz (commented):
@yzhao321 
one question: we have bind_policy and ISSUE_QUEUE_FU_EARLY_BIND = on/off

wouldn't it make more sense to remove ISSUE_QUEUE_FU_EARLY_BIND and have an empty bind policy? 

because for the other policy (sched/traversal) we do not have the ON/OFF option either.

hlitz (approved):
LGTM

PR #522, opened 2026-05-29
[FEATURE] Impl of AMD Bulldozer Random Sched Policy

yzhao321:
This patch introduces AMD Bulldozer random scheduling policy. It prioritizes one oldest op if ready, and picks ready ops in physical order.
No perf drift when the policy is not selected.


hlitz (approved):
LGTM, can you show perf results?

PR #525, opened 2026-06-04
[FEATURE] Priority Traversal Policy

yzhao321:
No perf drift if the policy is not selected


src/core.param.def
```diff
@@ -156,6 +156,8 @@ DEF_PARAM(issue_queue_schedule_policy, ISSUE_QUEUE_SCHEDULE_POLICY, uns, uns, 0,
 DEF_PARAM(issue_queue_traversal_policy, ISSUE_QUEUE_TRAVERSAL_POLICY, uns, uns, 0, )
 DEF_PARAM(issue_queue_early_bind_policy, ISSUE_QUEUE_EARLY_BIND_POLICY, uns, uns, 1, )
 
+DEF_PARAM(issue_queue_traversal_priority, ISSUE_QUEUE_TRAVERSAL_PRIORITY, char*, string, "0,1,2,3,4;-;-", )
```
hlitz, line 159: what if we have multiple issue queues? Like for LD/ST we have fewer FUs then for ALU and we might need another order. It would be good to have all policies to be per IQ but I think here it is needed.

yzhao321: @hlitz 
multi-queue setting is allowed.
';' seperates different queues. '-' indecates RR. 
"0,1,2,3,4;-;-" means '01234' for first queue, RR for queue 2 queue3

hlitz (approved):
LGTM

PR #527, opened 2026-06-11
Add cache index hashing mechanism

estodi:


src/bp/bp_targ_mech.c
```diff
@@ -471,17 +485,19 @@ void bp_btb_gen_init(Bp_Data* bp_data, Bp_Data* primary_bp) {
     for (uns ii = 0; ii < BTB_BANKS; ii++) {
       char name[MAX_STR_LENGTH + 1];
       snprintf(name, MAX_STR_LENGTH, "BTB BANK %d", ii);
-      init_cache_impl(&bp_data->btb[ii], name, BTB_ENTRIES / BTB_BANKS, BTB_ASSOC, 1, sizeof(Addr), REPL_TRUE_LRU,
-                      BTB_HASH);
+      init_cache_impl(&bp_data->btb[ii], name, BTB_ENTRIES / BTB_BANKS, BTB_ASSOC, 1, BTB_TAG_BITS, sizeof(Addr),
+                      REPL_TRUE_LRU, BTB_HASH, BTB_NUM_TRACK_BITS, BTB_INTERVAL_SIZE, BTB_REMAP_RATE,
+                      BTB_SWITCH_THRESH_PCT, BTB_MAIN_EI_SWITCHES);
```
hlitz: If I understand correctly these parameters are Entropy-index specific and not used for other indexing mechanisms. Can we have a seperate index_mechansism struct which is passed into the init_cache_impl ?

Also within cache-lib we nee to properly abstract the index mechanism so the indexing does not pollute the whole cache-lib. E..g have some well defined API compute_index() that can be overloaded depending on the chosen indexing mechanism.

You need to tell claude what type of indexing mechanisms you are planning to implement so that it can build the right abstraction e.g.:
- static XOR
- multi-bit XOR
- skewed associative hashing
- entropy index

src/libs/cache_lib.c
```diff
@@ -74,37 +87,50 @@ char rand_repl_state[31];
...
-static inline uns cache_index(Cache* cache, Addr addr, Addr* tag, Addr* line_addr) {
+static inline uns cache_index(Cache* cache, Addr addr, Addr* tag, Addr* tag_full, Addr* line_addr) {
   if (cache->tag_incl_offset) {
-    *tag = addr & ~(cache->set_mask << cache->shift_bits);
+    *tag = *tag_full = addr & ~(cache->set_mask << cache->shift_bits);
     *line_addr = addr;  // When the tag incl offset, cache is BYTE-addressable
   } else {
-    *tag = addr >> cache->shift_bits & cache->tag_mask;
+    Addr _tag;
+    _tag = *tag_full = addr >> cache->shift_bits;
```
hlitz, line 81: why do we no longer need the mask?

src/libs/cache_lib.c
```diff
@@ -74,37 +87,50 @@ char rand_repl_state[31];
...
+void init_cache_impl(Cache* cache, const char* name, uns cache_size, uns assoc, uns line_size, uns tag_bits,
+                     uns data_size, Repl_Policy repl_policy, Index_Hash_Id index_hash_id, uns ei_num_track_bits,
+                     uns ei_interval_size, uns ei_remap_rate, int ei_switch_thresh_pct, int ei_stat_base) {
   uns num_lines = cache_size / line_size;
   uns num_sets = cache_size / line_size / assoc;
   uns ii, jj;
 
   DEBUG(0, "Initializing cache called '%s'.\n", name);
 
+  cache->entropy_state = NULL;
```
hlitz: abstract into cache->index_mechanism.init();

src/libs/cache_lib.c
```diff
@@ -122,11 +148,17 @@ void init_cache_impl(Cache* cache, const char* name, uns cache_size, uns assoc,
...
+  cache->tag_bits = tag_bits;
   cache->shift_bits = LOG2(line_size);                /* use for shift amt. */
   cache->set_mask = N_BIT_MASK(LOG2(num_sets));       /* use after shifting */
-  cache->tag_mask = ~cache->set_mask;                 /* use after shifting */
+  cache->tag_pure_mask = tag_bits == 64 ? N_BIT_MASK_64 : N_BIT_MASK(tag_bits);
+  cache->tag_mask = cache->tag_pure_mask << LOG2(num_sets); /* use after shifting */
   cache->offset_mask = N_BIT_MASK(cache->shift_bits); /* use before shifting */
 
+  /* set up EntropyIndex state if this cache uses the entropy index hash */
+  if (index_hash_id == ENTROPY_INDEX)
```
hlitz: no index specific checks in the code, needs to be abstracted through some API

hlitz (changes requested):

src/libs/cache_lib.c
estodi: This is to include the set index bits in tag_full. If a hash function maps two distinct addresses that have exactly the same (old, partial) tag bits to a single cache set, there is no way to distinguish them without using those original set bits. So we need to include them in tag_full as well.

hlitz (approved):
LGTM

PR #529, opened 2026-06-11
EIP: TC 2024 squash-based wrong-path handling

5surim:
- Implements the wrong-path handling approach from TC 2024 (Ros & Jimborean, "Wrong-Path-Aware Entangling Instruction Prefetcher").
- Without this fix, speculative (wrong-path) accesses pollute EIP's entanglement learning — history buffer entries and timing MSHR entries created during wrong-path execution persist after recovery, causing EIP to learn incorrect entanglements and underperform.
- The fix tags all history and timing entries with instruction IDs (op_num) and squashes entries beyond the recovery point on misprediction recovery. This is implementable in real hardware (unlike gating on the simulator's oracle off_path flag).

Changes

- eip.cc:
  - Add op_num field to l1i_hist_entry and l1i_timing_mshr_entry structs
  - Stamp op_num on all new entries in l1i_add_hist_table() and l1i_add_timing_entry()
  - eip_prefetch() takes new op_num parameter, saves as l1i_current_op_num
  - New eip_recover(): invalidates history/timing entries with op_num > recovery_op_num, resets BB tracking state, saves recovery point for fill-time filtering
  - eip_cache_fill(): skips entanglement creation if the MSHR entry's op_num > l1i_recovery_op_num
- eip.h: Updated eip_prefetch() signature, added eip_recover() declaration
- icache_stage.c: Passes op_count[proc_id] to eip_prefetch(), calls eip_recover() in recover_icache_stage()


src/prefetcher/eip.cc
```diff
@@ -1166,10 +1204,15 @@ void eip_cache_fill(uns proc_id, uint64_t v_addr, uint64_t evicted_v_addr) {
...
 
   uint32_t pos_hist = L1I_HIST_TABLE_ENTRIES;
-  uint64_t latency = l1i_get_latency_timing_mshr(line_addr, pos_hist);
+  uint64_t mshr_op_num = 0;
+  uint64_t latency = l1i_get_latency_timing_mshr(line_addr, pos_hist, mshr_op_num);
 
   l1i_move_timing_entry(line_addr);
 
+  // TC'24: skip entanglement creation if MSHR entry was from wrong-path
+  if (mshr_op_num > l1i_recovery_op_num)
```
hlitz: will this not gate all future updates too (those that are created after flush) because new (mshr_)op_num will also be greater than recovery_op_nm?

hlitz (commented):
It looks good. Can you provide Spec17 numbers for FDIP vs. FDIP+old_EIP vs FDIP+new_EIP?

src/prefetcher/eip.cc
```diff
@@ -367,7 +368,9 @@ uint32_t l1i_add_hist_table(uint64_t line_addr) {
 
 void l1i_add_bb_size_hist_table(uint64_t line_addr, uint32_t bb_size) {
   uint64_t index = l1i_find_hist_entry(line_addr);
-  l1i_hist_table[eip_proc_id][index].bb_size = bb_size & L1I_MERGE_BBSIZE_MAX_VALUE;
+  if (index < L1I_HIST_TABLE_ENTRIES) {
```
hlitz, line 376: what happens if index is not smaller?

5surim: When `index >= L1I_HIST_TABLE_ENTRIES`, l1i_find_hist_entry did not locate line_addr in the 16-entry ring buffer. The entry we inserted for l1i_last_basic_block earlier can be evicted before we come back to annotate its bb_size: if the cycle gap is large, the time-overflow loop in
  l1i_add_hist_table (eip.cc:348-354) fills the ring with empty entries and pushes our entry out. In that case there is nothing to annotate, so skipping the write is the correct behavior.

  Note: the original ISCA-21 reference at `alberto-ros/EntanglingInstructionPrefetcher` writes at l1i_hist_table[index] unconditionally, which is a latent OOB write on the sentinel. The reference uses a static BSS array so the write is absorbed into adjacent globals; scarab's malloced
   per-core array exposes the bug (SIGSEGV under slurm's cgroup allocator). The bounds check makes the drop-the-annotation behavior explicit and memory-safe.

  I pushed `94816a2` adding a comment above the check to state the root cause in the source.

hlitz (approved):
LGTM

PR #530, opened 2026-06-17
[REFACTOR] Ready Not Issued PMU

yzhao321:
No perf drift.


src/issue_queue.cc
```diff
@@ -273,7 +274,10 @@ class SelectLogic {
 
   std::vector<size_t> picker_order;  // picker traversal order generated each cycle
   std::list<IssueQueueEntry*> ready_list;
-  uns64 ready_op_types = 0;  // bitmask of ready op types in this cycle
+
+  // bitmask of ready but not yet issued op types in this cycle
+  uns64 ready_not_issue_op_types = 0;
```
hlitz: issue->issued

src/issue_queue.cc
```diff
@@ -342,26 +349,48 @@ void SelectLogic::select() {
...
     FunctionalUnitPicker& fu_picker = connected_fu_pickers[picker_order[i]];
-    fu_picker.grant();
+    fu_picker.grant();  // TODO: let grant become certain pick by advance checking fu instead of rejecting in exec
 
     // track FU idle stats after scheduling
     uns32 fu_id = fu_picker.get_fu_id();
     if (node->sd.ops[fu_id] != NULL)
       continue;
 
+    if (ready_not_issued_op_types_others & fu_picker.get_fu_type()) {
```
hlitz, line 375: are these stats incremented only if the FU did not do any pick or always? ie. when an FU picks type A is it still incremented for type B? (assuming A and B was ready and the FU can pick A and B)

yzhao321: Two PMUs are introduced.
ISSUE_QUEUE_MATCHING_UNPICK_ACROSS_QUEUES tracks unpick ready across diff queues.
ISSUE_QUEUE_MATCHING_UNPICK_WITHIN_QUEUE tracks unpick ready within a queue.
Both increment only the FU does not pick.

hlitz (approved):
LGTM

I think we should discuss the cancel replay issue indepenently

PR #531, opened 2026-06-18
prefetcher: fix L2/UMLC prefetch feedback counters (sent, useful)

yql5510718:
**Summary**
L2-targeted (DEST_MLC) hardware prefetches had broken feedback accounting: the per-prefetcher useful signal was never delivered, and L2 prefetch sends were counted under the L1 stat. This left the L2 prefetcher's feedback counters dead/misattributed, even though the prefetcher itself worked. This PR fixes the two counter bugs.

**Background**
Prefetchers attach at UL1 (L1D) or UMLC (L2) via --pref_ul1_on / --pref_umlc_on. The UL1 feedback path is fully wired; the UMLC path was not:
pref_umlc_pref_hit() is defined, dispatched in pref_common.c, and registered in pref_table.def, but was never called from memory.c (only the UL1 equivalent pref_ul1_pref_hit() was), so curr_useful_core stayed 0 for any L2 prefetcher.
PREF_MLC_TOTAL_SENT was incremented nowhere. The bus-out send site called pref_ul1sent() unconditionally, so L2 prefetch sends were counted as PREF_L1_TOTAL_SENT and fed curr_sent_core under the wrong level.

**Changes**
Useful — wire pref_umlc_pref_hit() into the MLC prefetch-hit path in mem_process_mlc_hit_access(), mirroring the existing UL1 pref_ul1_pref_hit() call (fires once per useful line, next to PREF_MLC_TOTAL_USED).
Sent — add pref_umlcsent() (UMLC counterpart of pref_ul1sent(), emitting PREF_MLC_TOTAL_SENT / CORE_PREF_MLC_SENT) and route the bus-out send by destination.

The shared counter-bump + PREF_DHAL degree-adjust logic (dest-agnostic) is factored into a static pref_sent_update_degree() helper; pref_ul1sent is refactored to call it, so its behavior is unchanged.

bwaves_r/26004, stride prefetcher @ MLC, golden_cove:
Stat | Before | After
-- | -- | --
PREF_MLC_TOTAL_SENT | 0 | 1,057,742
PREF_L1_TOTAL_SENT (mislabel) | 1,058,001 | 259
UL1-config sanity (L1 / MLC sent) | 1.08M / 0 | unchanged
IPC | 0.82367 | 0.82367


src/memory/memory.c
```diff
@@ -1596,7 +1602,12 @@ static Flag mem_complete_l1_access(Mem_Req* req, Mem_Queue_Entry* l1_queue_entry
            req->type == MRT_FDIPPRFOFF || req->demand_match_prefetch) &&
           req->prefetcher_id != 0) {  // cmp FIXME What can I do for the prefetcher?
 
-        pref_ul1sent(req->proc_id, req->addr, req->prefetcher_id);
+        // Account the send against the cache level this prefetch targets so the
+        // per-prefetcher sent counters (PREF_*_TOTAL_SENT) track the right level.
+        if (req->destination == DEST_MLC)
+          pref_umlcsent(req->proc_id, req->addr, req->prefetcher_id);
+        else
```
hlitz: what if dest==dcache ? 
why do we not have 3 possible targets here?

src/prefetcher/pref_common.c
```diff
@@ -856,6 +847,42 @@ void pref_ul1sent(uns8 proc_id, Addr addr, uns8 prefetcher_id) {
...
+
+  // prefetch missed in the ul1 and went out on the bus
+  STAT_EVENT_ALL(PREF_L1_TOTAL_SENT);
+  STAT_EVENT(proc_id, CORE_PREF_L1_SENT);
+  pref_sent_update_degree(proc_id, prefetcher_id);
+}
+
+// UMLC counterpart of pref_ul1sent for prefetches issued toward the MLC/L2
+// (DEST_MLC): records the send under the L2 sent counters.
+void pref_umlcsent(uns8 proc_id, Addr addr, uns8 prefetcher_id) {
```
hlitz: a lot of replicated code here, maybe pass the req/target to update the correct counters?

hlitz (changes requested):

src/memory/memory.c
yql5510718: The dcache/DL0 path doesn't issue its own memory request — it probes the dcache and, on miss, forwards into the UL1 queue at pref_common.c:703, so it becomes a DEST_L1 send. No DEST_DCACHE send ever reaches req.

src/prefetcher/pref_common.c
```diff
@@ -856,6 +847,37 @@ void pref_ul1sent(uns8 proc_id, Addr addr, uns8 prefetcher_id) {
...
+    // UPDATE prefpolbv reset entry
+    pref_polbv_update_on_repref(proc_id, addr);
+  }
+
+  switch (dest) {
+    case DEST_MLC:
+      STAT_EVENT_ALL(PREF_MLC_TOTAL_SENT);
+      STAT_EVENT(proc_id, CORE_PREF_MLC_SENT);
+      break;
+    default:  // DEST_L1, and demand-matched prefetches whose dest was demoted
```
hlitz: I thought we had agreed on not using default for incrementing any counters but instead check explicitly. It is really bad coding practice.

For almost every switch/case you want to use default only to capture errors/unexpected cases.

Imagine someone adds a new destination then it will just count inccorrectly again. You cannot expect a developer to know all the code and all the switches distributed across the code base. If you assert in the default case then a new dest immediatelly triggers errors everywhere and it is clear what needs to be fixed.

hlitz (changes requested):
please always fix changes using another commit instead of amend/force.
1. I need to re-review ALL code not only the new patches
2. Previous comments are no longer visible so it is difficult to see what comments where addressed

hlitz (approved):
LGTM

PR #541, opened 2026-06-30
prefetcher: add cache_fill feedback hooks and opt-in dl0-fill-to-dcache

yql5510718:
Two general, opt-in prefetch-framework hooks. Both are additive and gated, so existing prefetchers (stream/stride/ghb/2dc/markov/...) are unchanged.

1. cache_fill dispatch. New `ul1_cache_fill` / `umlc_cache_fill` HWP callbacks plus `pref_ul1_cache_fill` / `pref_umlc_cache_fill` dispatchers, invoked at the LLC fill (l1_fill_line) and the L2 fill (mlc_fill_line). Mirrors ChampSim's *_prefetcher_cache_fill so eviction/fill-trained prefetchers and per-prefetcher accuracy accounting can observe fills. Fields are kept LAST in HWP so existing pref_table.def rows zero-initialize them.

2. dl0-fill-to-dcache. New --pref_dl0_fill_dcache (default FALSE). When set, a DL0 (L1D) prefetch issues a real memory request that fills the dcache on return (done_func = dcache_fill_line, dest = DEST_DCACHE) instead of forwarding it to the UL1 (LLC) request queue, so L1D-targeted prefetchers actually populate L1D. Default off preserves the legacy forward-to-UL1 behavior.


hlitz (approved):
LGTM

but please remember. don't force push but instead just add another commit on top. Then i would not have had to review everything from scratch again

PR #543, opened 2026-07-03
prefetcher: configurable per-level prefetcher instances and destinations

yql5510718:
**Summary**
Makes the prefetch framework instance-based: each prefetcher instance is declared at a training level (which cache's hits/misses it observes) and carries its own destination level (where its prefetches go). Any number of prefetchers per level, the same type at multiple levels, and dcache (L1D) as a first-class destination.

**Configuration** (runtime params, replaces pref_<type>_on / pref_umlc_on / pref_ul1_on)

--pref_num_dcache_prefetchers 2
--pref_dcache_prefetchers TYPE_STRIDE,DEST_DCACHE,TYPE_STREAM,DEST_L1
--pref_num_mlc_prefetchers 1
--pref_mlc_prefetchers    TYPE_STRIDE,DEST_MLC
--pref_num_l1_prefetchers 1
--pref_l1_prefetchers     TYPE_STRIDE,DEST_L1
The list fixes the training level; the DEST_* token fixes the destination — fully decoupled.
Being listed enables a type; pref_num_* = 0 disables a level; mismatches, unknown tokens, and duplicate types assert at init.
Routing goes through one choke point (pref_addto_dest_req_queue) on the activated HWP_Type; all six legacy prefetchers gain dcache-trained instances; new PREF_DCACHE_* stat family; in-repo PARAMS files migrated behavior-preservingly.
**Verification**
Bit-identical to main on bwaves_r ×3 simpoints (default sunny_cove and explicit ul1/umlc configs): cycles and every cumulative prefetch counter equal — table in comments. Multi-level example runs ASan-clean; config asserts verified in dbg.


src/prefetcher/pref_stride.c
```diff
@@ -77,12 +77,12 @@ void pref_stride_init(HWP* hwp) {
 
   if (PREF_UMLC_ON) {
     stride_prefetche_array.stride_hwp_umlc = (Pref_Stride*)malloc(sizeof(Pref_Stride));
-    stride_prefetche_array.stride_hwp_umlc->type = UMLC;
+    stride_prefetche_array.stride_hwp_umlc->type = PREF_TO_UMLC;
```
hlitz: similarly as above:
1. are we sure that its ok to enable UL1 and MLC prefetch? Are we sure that no state would be shared between such 2 prefetcher instances and that they are treated independently?
2. is it even required (because of inclusivity?)
3. If not, we need to assert that only one is enabled and we do notneed the if (Just set the type)
4. I suggest adding to all prefetchers the L1D case

hlitz (changes requested):

src/prefetcher/pref_2dc.c
```diff
@@ -83,12 +83,12 @@ void pref_2dc_init(HWP* hwp) {
 
   if (PREF_UMLC_ON) {
     tdc_prefetcher_array.tdc_hwp_umlc = (Pref_2DC*)malloc(sizeof(Pref_2DC));
-    tdc_prefetcher_array.tdc_hwp_umlc->type = UMLC;
+    tdc_prefetcher_array.tdc_hwp_umlc->type = PREF_TO_UMLC;
```
hlitz: 1. Why is there no case for L1D ?
2. Is it possible that both PREF_TO_UMLC and PREF_TO_UL1 is true? Why? Doesnt a prefetch to MLC automatically prefetch into UL1 because of inclusing cache hierarchy?

src/prefetcher/pref_common.c
```diff
@@ -133,6 +133,11 @@ void pref_init(void) {
   if (!PREF_FRAMEWORK_ON)
     return;
 
+  // A prefetcher targets a single destination level: never enable both.
+  // (Each enabled prefetcher's init additionally asserts at least one is set.)
+  ASSERTM(0, !(PREF_UMLC_ON && PREF_UL1_ON),
+          "enable at most one prefetch destination level (--pref_umlc_on / --pref_ul1_on)\n");
```
hlitz: I don't think this is correct. PREF_UMLC_ON and PREF_UL1_ON is independent of the target right?
I think you want to check for 
int num_pref = PREF_TO_DL0 + PREF_TO_UMLC + PREF_TO_UL1;
  ASSERTM(0, num_pref < 2);

src/prefetcher/pref_2dc.c
```diff
@@ -81,15 +81,21 @@ void pref_2dc_init(HWP* hwp) {
...
-  if (PREF_UMLC_ON) {
-    tdc_prefetcher_array.tdc_hwp_umlc = (Pref_2DC*)malloc(sizeof(Pref_2DC));
-    tdc_prefetcher_array.tdc_hwp_umlc->type = UMLC;
-    init_2dc(hwp, tdc_prefetcher_array.tdc_hwp_umlc);
-  }
-  if (PREF_UL1_ON) {
-    tdc_prefetcher_array.tdc_hwp_ul1 = (Pref_2DC*)malloc(sizeof(Pref_2DC));
-    tdc_prefetcher_array.tdc_hwp_ul1->type = UL1;
-    init_2dc(hwp, tdc_prefetcher_array.tdc_hwp_ul1);
+  // An enabled prefetcher needs a destination: at least one level must be on
```
hlitz: PREF_UMLC_ON and PREF_UL1_ON are about observing misses/hits and storing meta data and not about the destination, so this comment is confusing.

src/prefetcher/pref_2dc.c
```diff
@@ -81,15 +81,21 @@ void pref_2dc_init(HWP* hwp) {
...
+  // (pref_init asserts at most one, so exactly one twin is allocated here).
+  ASSERTM(0, PREF_UMLC_ON || PREF_UL1_ON,
+          "2dc: enable at least one destination level (--pref_umlc_on / --pref_ul1_on)\n");
+  switch ((PREF_UMLC_ON ? 1u : 0u) | (PREF_UL1_ON ? 2u : 0u)) {
```
hlitz: I don't think the switch case here improves readability, can we revert?

src/prefetcher/pref_2dc.c
```diff
@@ -81,15 +81,21 @@ void pref_2dc_init(HWP* hwp) {
...
-    tdc_prefetcher_array.tdc_hwp_umlc->type = UMLC;
-    init_2dc(hwp, tdc_prefetcher_array.tdc_hwp_umlc);
-  }
-  if (PREF_UL1_ON) {
-    tdc_prefetcher_array.tdc_hwp_ul1 = (Pref_2DC*)malloc(sizeof(Pref_2DC));
-    tdc_prefetcher_array.tdc_hwp_ul1->type = UL1;
-    init_2dc(hwp, tdc_prefetcher_array.tdc_hwp_ul1);
+  // An enabled prefetcher needs a destination: at least one level must be on
+  // (pref_init asserts at most one, so exactly one twin is allocated here).
+  ASSERTM(0, PREF_UMLC_ON || PREF_UL1_ON,
```
hlitz: you want to check for:  ASSERTM(0, !PREF_DCACHE_ON) or implement the APIs/initialization required for supporting DCACHE.

src/prefetcher/pref_ghb.c
```diff
@@ -77,17 +77,23 @@ void pref_ghb_init(HWP* hwp) {
...
-    for (uns i = 0; i < NUM_CORES; i++)
-      ghb_prefetchers_array.ghb_hwp_core_umlc[i].type = UMLC;
-    init_ghb_core(hwp, ghb_prefetchers_array.ghb_hwp_core_umlc);
-  }
-  if (PREF_UL1_ON) {
-    ghb_prefetchers_array.ghb_hwp_core_ul1 = (Pref_GHB*)calloc(NUM_CORES, sizeof(Pref_GHB));
-    for (uns i = 0; i < NUM_CORES; i++)
-      ghb_prefetchers_array.ghb_hwp_core_ul1[i].type = UL1;
-    init_ghb_core(hwp, ghb_prefetchers_array.ghb_hwp_core_ul1);
+  // An enabled prefetcher needs a destination: at least one level must be on
```
hlitz: revert switch/case

src/prefetcher/pref_markov.c
```diff
@@ -72,19 +72,26 @@ void pref_markov_init(HWP* hwp) {
...
-    markov_prefetchers_array.last_miss_addr_core_umlc = malloc(sizeof(Addr) * NUM_CORES);
-    init_markov(hwp, markov_prefetchers_array.markov_hwp_core_umlc, markov_prefetchers_array.last_miss_addr_core_umlc);
-  }
-  if (PREF_UL1_ON) {
-    markov_prefetchers_array.markov_hwp_core_ul1 = (Pref_Markov*)calloc(NUM_CORES, sizeof(Pref_Markov));
-    for (uns i = 0; i < NUM_CORES; i++)
-      markov_prefetchers_array.markov_hwp_core_ul1[i].type = UL1;
-    markov_prefetchers_array.last_miss_addr_core_ul1 = malloc(sizeof(Addr) * NUM_CORES);
-    init_markov(hwp, markov_prefetchers_array.markov_hwp_core_ul1, markov_prefetchers_array.last_miss_addr_core_ul1);
+  // An enabled prefetcher needs a destination: at least one level must be on
```
hlitz: revert switch/case

src/prefetcher/pref_stream.c
```diff
@@ -103,13 +103,19 @@ void pref_stream_init(HWP* hwp) {
...
-  if (PREF_UMLC_ON) {
-    stream_prefetchers_array.pref_stream_core_umlc = (Pref_Stream*)calloc(NUM_CORES, sizeof(Pref_Stream));
-    init_stream_core(hwp, stream_prefetchers_array.pref_stream_core_umlc);
-  }
-  if (PREF_UL1_ON) {
-    stream_prefetchers_array.pref_stream_core_ul1 = (Pref_Stream*)calloc(NUM_CORES, sizeof(Pref_Stream));
-    init_stream_core(hwp, stream_prefetchers_array.pref_stream_core_ul1);
+  // An enabled prefetcher needs a destination: at least one level must be on
+  // (pref_init asserts at most one, so exactly one twin is allocated here).
+  ASSERTM(0, PREF_UMLC_ON || PREF_UL1_ON,
```
hlitz: revert switch/case

src/prefetcher/pref_stride.c
```diff
@@ -75,15 +75,21 @@ void pref_stride_init(HWP* hwp) {
...
-  if (PREF_UMLC_ON) {
-    stride_prefetche_array.stride_hwp_umlc = (Pref_Stride*)malloc(sizeof(Pref_Stride));
-    stride_prefetche_array.stride_hwp_umlc->type = UMLC;
-    init_stride(hwp, stride_prefetche_array.stride_hwp_umlc);
-  }
-  if (PREF_UL1_ON) {
-    stride_prefetche_array.stride_hwp_ul1 = (Pref_Stride*)malloc(sizeof(Pref_Stride));
-    stride_prefetche_array.stride_hwp_ul1->type = UL1;
-    init_stride(hwp, stride_prefetche_array.stride_hwp_ul1);
+  // An enabled prefetcher needs a destination: at least one level must be on
```
hlitz: revert switch/case

src/prefetcher/pref_stridepc.c
```diff
@@ -72,17 +72,23 @@ void pref_stridepc_init(HWP* hwp) {
...
-    for (uns i = 0; i < NUM_CORES; i++)
-      stridepc_prefetche_array.stridepc_hwp_core_umlc[i].type = UMLC;
-    init_stridepc(hwp, stridepc_prefetche_array.stridepc_hwp_core_umlc);
-  }
-  if (PREF_UL1_ON) {
-    stridepc_prefetche_array.stridepc_hwp_core_ul1 = (Pref_StridePC*)calloc(NUM_CORES, sizeof(Pref_StridePC));
-    for (uns i = 0; i < NUM_CORES; i++)
-      stridepc_prefetche_array.stridepc_hwp_core_ul1[i].type = UL1;
-    init_stridepc(hwp, stridepc_prefetche_array.stridepc_hwp_core_ul1);
+  // An enabled prefetcher needs a destination: at least one level must be on
```
hlitz: revert switch/case

src/prefetcher/pref_2dc.c
```diff
@@ -81,15 +81,21 @@ void pref_2dc_init(HWP* hwp) {
...
+          "2dc: enable at least one destination level (--pref_umlc_on / --pref_ul1_on)\n");
+  switch ((PREF_UMLC_ON ? 1u : 0u) | (PREF_UL1_ON ? 2u : 0u)) {
+    case 1u:
+      tdc_prefetcher_array.tdc_hwp_umlc = (Pref_2DC*)malloc(sizeof(Pref_2DC));
+      tdc_prefetcher_array.tdc_hwp_umlc->type = PREF_TO_UMLC;
```
hlitz: No, we said the destination is taken from pref_table.def Didn't we add this in an earlier version?

The prefetch destination is independent from the trigger (miss/hit observation point)

hlitz (changes requested):

src/memory/memory.stat.def
```diff
@@ -1408,6 +1408,8 @@ DEF_STAT(  PREF_MLC_TOTAL_PARTIAL_USED   , COUNT , NO_RATIO  )
 DEF_STAT(  PREF_MLC_TOTAL_NOT_USED       , COUNT , NO_RATIO  )
 DEF_STAT(  PREF_MLC_TOTAL_DROPPED        , DIST  , NO_RATIO  )
 
+DEF_STAT(  PREF_DCACHE_TOTAL_SENT	        , COUNT , NO_RATIO  )
```
hlitz, line 1411: why not replicate the other stats above (used,partial, dropped) ?

src/prefetcher/pref_common.c
```diff
@@ -921,12 +960,16 @@ void pref_sent(uns8 proc_id, Addr addr, uns8 prefetcher_id, Destination dest) {
       STAT_EVENT_ALL(PREF_MLC_TOTAL_SENT);
       STAT_EVENT(proc_id, CORE_PREF_MLC_SENT);
       break;
+    case DEST_DCACHE:  // dl0 prefetch filling the dcache directly (--pref_dl0_fill_dcache)
```
hlitz: why are we using 2 terms for the same thing? e.g. PREF_TO_DL0 vs DEST_DCACHE? 

I prefer DCACHE over DL0

src/prefetcher/pref.param.def
```diff
@@ -53,6 +53,16 @@ DEF_PARAM( pref_framework_on                   , PREF_FRAMEWORK_ON
 DEF_PARAM( pref_trace_on                       , PREF_TRACE_ON                       , Flag            , Flag               , FALSE     ,    )
 DEF_PARAM( pref_umlc_on                        , PREF_UMLC_ON                         , Flag            , Flag               , FALSE     ,    )
 DEF_PARAM( pref_ul1_on                         , PREF_UL1_ON                         , Flag            , Flag               , TRUE      ,    )
+/* Per-prefetcher destination level override: 0 = UL1 (LLC), 1 = UMLC (L2),
+   2 = DL0 (L1D) -- matching the HWP_Type enum; 32 = use the prefetcher's
+   pref_table dest_level. All of a prefetcher's prefetches go to this one level,
+   regardless of which cache level's hooks trained it. */
+DEF_PARAM( pref_stream_dest_level              , PREF_STREAM_DEST_LEVEL              , uns             , uns                , 32        ,    )
```
hlitz: why do we need two locations to define the dest level. e.g. pref.param.def and pref_table.def.

In the previous patch we had no option to define dest now we have 2. Why is one insufficient?

hlitz (changes requested):

src/prefetcher/pref.param.def
```diff
@@ -53,16 +53,16 @@ DEF_PARAM( pref_framework_on                   , PREF_FRAMEWORK_ON
...
-DEF_PARAM( pref_stride_dest_level              , PREF_STRIDE_DEST_LEVEL              , uns             , uns                , 32        ,    )
-DEF_PARAM( pref_stridepc_dest_level            , PREF_STRIDEPC_DEST_LEVEL            , uns             , uns                , 32        ,    )
-DEF_PARAM( pref_2dc_dest_level                 , PREF_2DC_DEST_LEVEL                 , uns             , uns                , 32        ,    )
-DEF_PARAM( pref_ghb_dest_level                 , PREF_GHB_DEST_LEVEL                 , uns             , uns                , 32        ,    )
-DEF_PARAM( pref_markov_dest_level              , PREF_MARKOV_DEST_LEVEL              , uns             , uns                , 32        ,    )
+/* Per-prefetcher destination cache level: 0 = UL1 (LLC), 1 = UMLC (L2),
+   2 = DCACHE (L1D) -- matching the HWP_Type enum. The single place a
+   prefetcher's destination is defined. All of a prefetcher's prefetches go to
+   this one level, regardless of which cache level's hooks trained it. */
+DEF_PARAM( pref_stream_dest_level              , PREF_STREAM_DEST_LEVEL              , uns             , uns                , 0         ,    )
```
hlitz: why do we need this here instead of configuring all prefetcher stuff in one place? Wouldn't it be more natural to remove the additional params here and define evertying in pref_table_def?

yql5510718: The table is compiled into the binary, so a table-only definition can't be changed at runtime. Even if we register api in table still need argument to assign the destination. Params are Scarab's only post-compile configuration mechanism.

How this evolved across the commits:

When the PR is initially posted,  the destination was hard-coded in each pref_X.c (each twin branched if (type==UMLC) pref_addto_umlc_req_queue else pref_addto_ul1req_queue at every emit site). Fixed at compile time, and entangled with the training level.
Previous commit — pref_table.def held a compile-time default, and the param overrode it. That was two places defining the same thing.
This commit — only the param remains. We kept the param rather than the table because the table can't do the job alone: selecting stride→L2 vs stride→LLC for an experiment would mean editing pref_table.def and rebuilding, and mixed-level runs (stream→LLC + stride→L2 in one binary) or level sweeps would each need their own build.
There's also a consistency argument: pref_table.def wires code (init/hook function pointers), while every runtime tunable a prefetcher has (pref_stream_on, degrees, table sizes) already lives in pref.param.def — so the destination level, being a tunable, sits next to them.

If compile-time-only configuration is acceptable, we can move it back into the table, but we couldn't find a way to configure the destination after compilation without a param.

hlitz: ok. can we do the following in the params file (dynamically configurable).

NUM_DCACHE_PREFETCHERS=2
NUM_MLC_PREFETCHERS=1
NUM_L1_PREFETCHERS=1
DCACHE_PREFETCHERS=TYPE_STRIDE,DEST_DCACHE,TYPE_STREAM,DEST_L1
MLC_PREFETCHERS=TYPE_STRIDE,DEST_MLC
L1_PREFETCHERS=TYPE_STRIDE,DEST_L1


For that APIs I am ok for now to have them statically configured although it would be better to have a single undified interface for all levels (DC,L1,MLC) and then just call the API on a specific instance.

hlitz (approved):
LGTM

PR #544, opened 2026-07-04
topdown corebound extensions

sweet12prof:
## Summary
Decomposes the CoreBound metric into two granular sub-components to better isolate pipeline starvation root causes: Core ILP Bound (lack of application parallelism) and Core Bandwidth Bound (structural/resource constraints).

## Metric Breakdown
* **Core ILP Bound**: Tracks empty slots when issue width > ready insts.
* **Core Bandwidth Bound**: Tracks ready instructions that couldn't issue due to resource/port contention.


src/issue_queue.cc
```diff
@@ -913,10 +916,15 @@ void IssueQueues::schedule() {
   update_mem_block();
 
   std::vector<uns64> ready_not_issued_op_types_total;
+  num_of_ready_ops = 0;
```
hlitz: Global variables read and written by different objects/classes is not good design practice.

you can:
1. introduce queue.ready_ops which count the per q ready ops
2. have a local stack variable in schedule which aggregates ready_ops accross queues

src/issue_queue.cc
```diff
@@ -913,10 +916,15 @@ void IssueQueues::schedule() {
...
   for (IssueQueue& queue : issue_queues) {
     queue.bid();
     ready_not_issued_op_types_total.push_back(queue.get_ready_not_issued_op_types());
   }
+  uns unused_slots = ISSUE_WIDTH - num_of_ready_ops;
+  if (unused_slots < 0)
```
hlitz: how can an unsigned (uns) ever be negative?
you should check for num_of_ready_ops > ISSUE_WIDTH
always check the inputs (ready_ops, issue_width) and never the result (unused_slots) to avoid overflow issues etc.

hlitz: rename to unused_ilp_slots

hlitz (changes requested):

src/issue_queue.cc
sweet12prof: Done

sweet12prof: Done

src/issue_queue.cc
sweet12prof: Done. Thank you for this insight

src/issue_queue.cc
```diff
@@ -319,13 +321,17 @@ void SelectLogic::bid() {
   traversal_policy->build_picker_order(picker_order);
   ready_not_issued_op_types = 0;
   ready_not_issued_op_types_per_fu.assign(connected_fu_pickers.size(), 0);
+  this->num_of_ready_ops = 0;
```
hlitz: NIT: remove this->

src/issue_queue.cc
```diff
@@ -319,13 +321,17 @@ void SelectLogic::bid() {
   traversal_policy->build_picker_order(picker_order);
   ready_not_issued_op_types = 0;
```
hlitz, line 324: we should probably be consistent with all PMUs here.

@yzhao321 should we always only count on-path? Are all other Topdown counters counting both on-path and off-path instructions? I feel only counting on-path may make sense because off-path is already accounted for as bad-spec?

hlitz (changes requested):

src/issue_queue.cc
sweet12prof: Done.

src/issue_queue.cc
hlitz: @yzhao321 can you comment on this? are all other topdown specific metrics on-path only?

yzhao321 (changes requested):
Hi @hlitz @sweet12prof 
Thanks and sorry for the delay.
I have several question.
1. For ISSUE_WIDTH: I think the current scarab ISSUE_WIDTH is incorrect. The current ISSUE_WIDTH is DISPATCH_WIDTH. In the golden cove, the current ISSUE_WIDTH config is 6, but i think it should be 5 (int/vec) + 3 (ld) + 4 (st, 2 in scarab). So should the ilb_unused_slot use this num to calculate instead of ISSUE_WIDTH? And do we need an extra patch to redefine the WIDTH?
2. For mem_bound, I think we may use all loads instead of on-path loads. In the paper Top-Down, the definition is 'any load'. In https://github.com/andikleen/pmu-tools/blob/master/skl_client_ratios.py, their mem stall don't use load_retire but use any.

src/issue_queue.cc
```diff
@@ -276,6 +276,8 @@ class SelectLogic {
   std::list<IssueQueueEntry*> ready_list;
 
   // bitmask of ready but not yet issued op types in this cycle
+  uns num_of_ready_ops_onpath = 0;
```
hlitz: can we shorten by removing "_of"

src/issue_queue.cc
```diff
@@ -276,6 +276,8 @@ class SelectLogic {
...
+  uns num_of_ready_ops_general = 0;
```
hlitz: can you rename to num_ready_ops_offpath as this is the common scarab convention

(we can add on+off to get the total)

src/issue_queue.cc
```diff
@@ -913,10 +926,25 @@ void IssueQueues::schedule() {
   update_mem_block();
 
   std::vector<uns64> ready_not_issued_op_types_total;
+  uns num_of_ready_ops_onpath = 0;
+  uns num_of_ready_ops_general = 0;
   for (IssueQueue& queue : issue_queues) {
     queue.bid();
     ready_not_issued_op_types_total.push_back(queue.get_ready_not_issued_op_types());
+    num_of_ready_ops_onpath += queue.get_num_of_ready_ops_onpath();
```
hlitz: -> num_ready_ops_total_onpath

src/issue_queue.cc
```diff
@@ -913,10 +926,25 @@ void IssueQueues::schedule() {
...
+    num_of_ready_ops_general += queue.get_num_of_ready_ops_general();
   }
+  uns unused_ilp_slots_onpath = DISPATCH_WIDTH - num_of_ready_ops_onpath;
```
hlitz: better do MIN(0, DISPATCH_WIDTH - num_of_ready_ops_onpath);

src/issue_queue.cc
```diff
@@ -913,10 +926,25 @@ void IssueQueues::schedule() {
...
+  uns unused_ilp_slots_general = DISPATCH_WIDTH - num_of_ready_ops_general;
+
+  if (num_of_ready_ops_onpath > DISPATCH_WIDTH)
+    unused_ilp_slots_onpath = 0;
+
+  if (num_of_ready_ops_general > DISPATCH_WIDTH)
+    unused_ilp_slots_general = 0;
+
+  INC_STAT_EVENT(proc_id, TOPDOWN_UNUSED_ILP_SLOTS_ONPATH, unused_ilp_slots_onpath);
+  INC_STAT_EVENT(proc_id, TOPDOWN_UNUSED_ILP_SLOTS_GENERAL, unused_ilp_slots_general);
```
hlitz: not sure what "general" means better use ALL

src/topdown.c
```diff
@@ -215,4 +215,29 @@ void topdown_done(uns proc_id) {
...
                  bad_spec_bound - GET_STAT_EVENT(proc_id, TOPDOWN_BR_RECOVER_AT_EXEC_BOUND));
+
+  uns64 unutilised_ilp_slots_onpath = GET_STAT_EVENT(proc_id, TOPDOWN_UNUSED_ILP_SLOTS_ONPATH);
+  uns64 unutilised_ilp_slots_general = GET_STAT_EVENT(proc_id, TOPDOWN_UNUSED_ILP_SLOTS_GENERAL);
+
+  uns64 unutilised_bandwidth_slots_onpath = GET_STAT_EVENT(proc_id, ST_OP_ONPATH_READY_NOT_ISSUED);
+  uns64 unutilised_bandwidth_slots_general =
+      GET_STAT_EVENT(proc_id, ST_OP_ONPATH_READY_NOT_ISSUED) + GET_STAT_EVENT(proc_id, ST_OP_OFFPATH_READY_NOT_ISSUED);
+
+  uns64 corebound_slots_onpath = unutilised_ilp_slots_onpath + unutilised_bandwidth_slots_onpath;
```
hlitz: not sure if corebound is the right term here. Better "wasted_slots_onpath"

src/topdown.c
```diff
@@ -215,4 +215,29 @@ void topdown_done(uns proc_id) {
...
+
+  uns64 corebound_slots_general = unutilised_ilp_slots_general + unutilised_bandwidth_slots_general;
+
+  INC_STAT_EVENT(proc_id, TOPDOWN_CORE_ILP_BOUND_ONPATH_CALCULATION,
+                 unutilised_ilp_slots_onpath * GET_STAT_EVENT(proc_id, TOPDOWN_CORE_BOUND) / corebound_slots_onpath);
+
+  INC_STAT_EVENT(proc_id, TOPDOWN_CORE_ILP_BOUND_GENERAL_CALCULATION,
+                 unutilised_ilp_slots_general * GET_STAT_EVENT(proc_id, TOPDOWN_CORE_BOUND) / corebound_slots_general);
+
+  INC_STAT_EVENT(proc_id, TOPDOWN_CORE_BANDWIDTH_BOUND_GENERAL_CALCULATION,
```
hlitz: would it be correct and easier to read if we used the same formula as above?

unutilised_bandwidth_slots_onpath * GET_STAT_EVENT(proc_id, TOPDOWN_CORE_BOUND) / corebound_slots_general);

hlitz (changes requested):

src/issue_queue.cc
sweet12prof: done

src/issue_queue.cc
sweet12prof: done

src/issue_queue.cc
sweet12prof: Sure. 
I couldnt find the MIN macro but i did find MIN2, MIN3 etc. 
Also, because the counters are uns, MIN2(0, DISPATCH_WIDTH - num_of_ready_ops_onpath) always returns 0 (bug). 

I did  DISPATCH_WIDTH - MIN2(DISPATCH_WIDTH, num_ready_ops...) instead.

src/issue_queue.cc
sweet12prof: Done

src/topdown.c
sweet12prof: Done

src/topdown.c
sweet12prof: Yes, I have done this accordingly.

src/issue_queue.cc
sweet12prof: done

sweet12prof:
With regards to calculating ONPATH ONLY  and ONPATH+OFFPATH topdown results so that we can compare both alternatives in scarab-infra, I am stuck on the ff issue. 

The first level's categories are expressed as fractions of TOPDOWN_TOTAL_SLOTS.
`  uns64 frontend_bound = GET_STAT_EVENT(proc_id, TOPDOWN_FETCH_BUBBLES_SLOTS) * TOPDOWN_SCALE_FACTOR /
                         GET_STAT_EVENT(proc_id, TOPDOWN_TOTAL_SLOTS);`
                         
`  uns64 bad_spec_bound = bad_spec_slots * TOPDOWN_SCALE_FACTOR / GET_STAT_EVENT(proc_id, TOPDOWN_TOTAL_SLOTS);`

`  uns64 retiring_bound = GET_STAT_EVENT(proc_id, TOPDOWN_RETIRED_SLOTS) * TOPDOWN_SCALE_FACTOR /
                         GET_STAT_EVENT(proc_id, TOPDOWN_TOTAL_SLOTS);`
                         
 `uns64 backend_bound = TOPDOWN_SCALE_FACTOR - frontend_bound - bad_spec_bound - retiring_bound;`
 
 
1. Estimating frontend-bound  with only ONPATH counters does not seem right, there will be the missing OFFPATH fraction for the frontend category that gets lumped into backend_bound through the last back calculation. 
 
2. If we had a slot calculation for backend_bound rather than the current back-calculation, For ONPATH_ONLY topdown calculation of backend_bound, the denominator of the fraction needs to remain TOPDOWN_TOTAL_SLOTS to be consistent with the bad_spec and retiring calculations. Putting together all the calculations, the total will not sum up to TOPDOWN_SCALE_FACTOR.

3. Making a  TOPDOWN_TOTAL_SLOTS_ONPATH does not seem right either since the bad spec category intrinsically consists of OFFPATH slots. 

I was thinking we leave category 1 calculations as is, and then have onpath vs offpath+onpath calculations and comparisons in downstream categories. The downstream categories are expressed as fractions of the preceding top level so everything should add up for either onpath only calculation or offpath+onpath calculation.

PR #545, opened 2026-07-04
Synthetic frontend

sweet12prof:
# Synthetic frontend 
Defines
-   Logic and Plugins for frontend 
-  APIs that generate microkernels that can be issued at the frontend

## Files 
Find files in src/frontend/synthetic
- sampler  :  builds a sequence (uniform, random, normal or discrete probabilities) with an api to sample elements. 
- synthetic_basic_apis :  low level apis that generates ctype_pin_insts of basic instructions .
- synthetic_kernels: consists of kernels generated using basic_apis. 
- kernel_table:  knobs that identify and can be used to select  between the synthetic kernels. 
- kernel_dispatcher: dispatches a selected kernel to the syntehtic frontend
- synth_fe : synthetic instruction frontend 

##  Usage
- add the following flags 
- ` .... --frontend synthetic --bottleneck  [ table_entry_from_bottleneck_table ] `

## Example Kernels Created
 "mem_bandwidth_limited"   -  sequence of independent loads
 "icache_limited"                      -  
 "dcache_limited"                     -   pointer chasing memory workload (limited to dcache)
 "mlc_limited"                           -   pointer chasing memory workload (all accesses miss dcache hit next level )
 "llc_limited"                              -   pointer chasing memory workload ( misses all levels but hits in llc )
 "mem_limited"                        -   pointer chasing memory workload ( misses all levels hits in main memory )
 "ilp_limited_0_dep_chain"     -   alu_type_workload, high ilp no dependendence
 "ilp_limited_1_dep_chain"     -   alu_type_workload, single loop carried dependendence chain
 "ilp_limited_2_dep_chain"     -  alu_type_workload, two independent but locally loop carried dependendence chains
 "ilp_limited_4_dep_chain"     -  alu_type_workload, four independent but locally loop carried dependendence chains
 "cbr_limited_20t"            -  CF workload, back-to-back cbr branches 20T/80NT. need issue_wodth-1 nops before every cbr
 "cbr_limited_50t"            -  CF workload, back-to-back cbr branches 50T/50NT. need issue_wodth-1 nops before every cbr
 "cbr_limited_80t"            -  CF workload, back-to-back cbr branches 80T/20NT. need issue_wodth-1 nops before every cbr
 "btb_limited_full_capacity_sweep - CF_BR workload, circles through the entire BTB entries + 1 100% miss
 "btb_limited_assoc_sweep" - does the same as full_capacity_sweep but using conflict misses
 "btb_contained" - this kernel exposes 100% BTB hit rate
 "ibr_limited_RR_4Tgts"  - cf_ibr workload, circles through 4 fixed indirect branch targets in a round robin fashion. indirect effective memory operand remains fixed 
 "ibr_limited_random_4Tgts" - cf_ibr workload, indirect target sampled randomly from 4 indirect target, indirect  effective memory address remains fixed
 "ibr_limited_random_2Tgts" - same as above but in this case only 2 random targets


src/frontend/synthetic/run.sh
```diff
@@ -0,0 +1,181 @@
+#!/bin/bash
```
hlitz, line 1: the functionality of this file needs to be moved into scarab-infra.

We should extend workloads_db.json with a new benchmark suite that lists all the new kernels and a new frontend type

src/frontend/synthetic/sampler.cc
```diff
@@ -0,0 +1,152 @@
...
+ */
+
+Sampler::Sampler(std::vector<uns64> discrete_values, Sequence_Pick_Strategy strategy, uns64 periodicity,
+                 uns sequence_length, std::vector<uns64> discrete_weights = {})
+    : strategy(strategy), next_pick_index(0) {
+  assert(!discrete_values.empty() && "sequence cannot be empty");
+  switch (strategy) {
+    case UNIFORM_SEQUENTIAL:
+    case USER_DEFINED:
+    default:  // <- default
```
hlitz: default always needs to be the last case.

Furthermore, what even is the expected default case? I prefer covering all cases expliclitly and in the default case have an assertion that it never happens.

src/frontend/synthetic/sampler.cc
```diff
@@ -0,0 +1,152 @@
...
+ * periodicity : number of sequence duplicates
+ * strategy    : suggests the pattern of the sequence
+ */
+
+Sampler::Sampler(Sequence_Pick_Strategy strategy, uns64 value_range, uns64 startval, uns64 stride, uns periodicity,
+                 std::vector<uns64> user_seq)
+    : strategy(strategy), next_pick_index(0) {
+  switch (strategy) {
+    case UNIFORM_SEQUENTIAL:
+    default:  // <- default
```
hlitz: default always needs to be the last case.

Furthermore, what even is the expected default case? I prefer covering all cases expliclitly and in the default case have an assertion that it never happens.

src/frontend/synthetic/sampler.cc
```diff
@@ -0,0 +1,152 @@
...
+ * sequence_length  : specifies length of the discrete distribution to be
+ generated.
+ * periodicity      : number of sequence duplicates
+ * strategy         : suggests the pattern of the sequence
+ NB: when strategy is UNIFORM_*, the discrete_values become the actual sequence
+ to be consumed, if NORMAL_RANDOM we generate a discrete sequence based on the
+ dicrete values and their corresponding weigths
+ */
+
+Sampler::Sampler(std::vector<uns64> discrete_values, Sequence_Pick_Strategy strategy, uns64 periodicity,
```
hlitz: why do we have 2 constructors with similar/partially redundant code? can we have the 2 constructors share the same code and they should only differ in the exact functionality added by one of the constructors

src/frontend/synthetic/sampler.cc
```diff
@@ -0,0 +1,152 @@
...
+  }
+}
+
+uns64 Sampler::get_next_element() {
+  auto index{next_pick_index};
+  if (++next_pick_index >= sequence_vector.size())
+    next_pick_index = 0;
+  return sequence_vector[index];
+}
+// seed
```
hlitz: why is the random dev defined as a global var? I think it should be a class member of the sampler

src/frontend/synthetic/sampler.cc
```diff
@@ -0,0 +1,152 @@
...
+  sequence_vector.reserve(sequence_vector.size() * periodicity);
+  for (uns i = 1; i < periodicity; i++) {
+    sequence_vector.insert(sequence_vector.end(), original_vector.begin(), original_vector.end());
+  }
+}
+
+uns64 Sampler::peek_element_following_next() const {
+  auto index = (next_pick_index + 1) % sequence_vector.size();
+  if (index < next_pick_index)
+    return sequence_vector.back();
```
hlitz: why do we return .back() and do not overflow? ie if next_pick_index points to the last element why does peek_element_following_next not return the first element of the vector?

src/frontend/synthetic/synthetic_basic_apis.cc
```diff
@@ -0,0 +1,109 @@
...
+  inst.inst_uid = uid;
+  inst.instruction_next_addr = tgt;
+  inst.size = inst_size;
+  inst.op_type = OP_CF;
+  inst.cf_type = CF_BR;
+  inst.num_simd_lanes = 1;
+  inst.lane_width_bytes = 1;
+  inst.branch_target = tgt;
+  inst.actually_taken = TAKEN;
+  strcpy(inst.pin_iclass, "DUMMY_UBR_JMP");
```
hlitz: can we explitly name them SYNTH so they are not confused with other dummy ops we have?

src/frontend/synthetic/synthetic_bottleneck_kernels.cc
```diff
@@ -0,0 +1,313 @@
...
+                             {taken_ratio, not_Taken_ratio});
+
+  std::map<uns64, ctype_pin_inst> kernel_map;
+  uns64 current_pc{start_pc}, _target{0};
+  ctype_pin_inst next_inst;
+
+  // Generate insts for every possible pc, includes possible offpath insts
+  for (uns i{0}; i < (2 * target_pool_size + PAD_LENGTH); i++) {
+    // generate leading nops and return next pc
+    current_pc = gen_issue_width_lock_nops(kernel_map, uid_sequence, ISSUE_WIDTH - 1, current_pc);
```
hlitz: what does the work "lock"mean in the API name?

src/frontend/synthetic/synthetic_bottleneck_kernels.cc
```diff
@@ -0,0 +1,313 @@
...
+  for (uns i{0}; i < num_of_nops; i++) {
+    auto current_pc{nop_pcs.get_next_element()};
+    kernel_map.insert({current_pc, generate_nop(current_pc, uid_sequence.get_next_element(), NOP_SIZE, false)});
+  }
+  return nop_pcs.get_next_element();
+}
+
+/*  Microkernel Definitions */
+
+// CBR
```
hlitz: add a description of each kernel a) describe the sequence generated b) describe what bottleneck the kernel stresses

src/frontend/synthetic/synthetic_bottleneck_kernels.cc
```diff
@@ -0,0 +1,313 @@
...
+  for (uns i{0}; i < ((2 * target_pool_size) + PAD_LENGTH); i++) {
+    // generate leading nops
+    current_pc = gen_issue_width_lock_nops(kernel_map, uid_sequence, ISSUE_WIDTH - 1, current_pc);
+    current_uid = uid_sequence.get_next_element();
+
+    // every taken target is 2 cachelines away
+    uns64 next_target = combined_target_pool.peek_element_following_next();
+
+    // if the generated target exceeds the taken targets distribution we set next target to beginning pc
+    if (next_target > targets_pool.get_last_element())
```
hlitz: why is this needed? I thought the sampler already implements a wrap around. i.e. if you use get_next it should always return a value that is in the distribution. It would be preferable to only get "legal" targets so that we do not have to check here (and in the other kernels) again

src/frontend/synthetic/synthetic_bottleneck_kernels.cc
```diff
@@ -0,0 +1,313 @@
...
+#include "frontend/synthetic/synthetic_kernels.h"
+#include "isa/isa.h"
+
+// pad length is the number of off-path insts to be padded to the backward branch at the tail end of every kernel
+#define PAD_LENGTH 300
+
+/* Helper Functions For Microkernels */
+
+// Function to generate leading nops for CF workloads
+uns64 gen_issue_width_lock_nops(std::map<uns64, ctype_pin_inst>& kernel_map, Sampler& uid_sequence, uns num_of_nops,
```
hlitz: Can you explain why the functions are defined as free functions and not member functions of a kernel object? I am not saying this design is wrong but I want to know your reasoning for taking this design.

This file defines some global variables that the functions use (  Sampler uid_sequence(UNIFORM_SEQUENTIAL, (workload_length + PAD_LENGTH), start_uid, 1, 1);). 

What is your reasoning for making these sampler variables global and not a member variable.  i could imagine that you want to mix multiple kernels and then you want to have a sampler per kernel

src/issue_queue.cc
```diff
@@ -326,6 +326,9 @@ void SelectLogic::bid() {
       continue;
     }
 
+    if (!entry->op->off_path)
```
hlitz, line 335: need to rebase to get rid of these changes here.

hlitz (changes requested):
In general this looks pretty good.

We need to come up with a plan to port run.sh into scarab-infra.

src/frontend/synthetic/sampler.cc
sweet12prof: Refactored.

src/frontend/synthetic/sampler.cc
sweet12prof: Refactored

src/frontend/synthetic/sampler.cc
sweet12prof: Done

src/frontend/synthetic/sampler.cc
sweet12prof: I did not wrap around in this particular function because the tail end of the kernel it is used in has a dedicated unconditional branch to the beginning. Also, it is used only by the CBR workload.

src/frontend/synthetic/synthetic_basic_apis.cc
sweet12prof: Done

src/issue_queue.cc
sweet12prof: Done

src/frontend/synthetic/kernel_factory.cc
```diff
@@ -0,0 +1,250 @@
...
+ctype_pin_inst Kernel_Factory::get_next_kernel_inst() {
+  auto it = kernel_map.find(get_next_pc());
+  auto inst = it->second;
+  assert(it != kernel_map.end() && "Every inst possible should be in the map");
+  // update next_pc for next time
+  next_pc = inst.instruction_next_addr;
+  return it->second;
+}
+
+// dipatches next cbr op and rerandomizes the cbr directions if we are at the end of kernel
```
hlitz: can you clarify why we are re-randomizing? If we do not want to re-execute the same sequence couldn't we just create a longer sequence? I think that if the number of simulated instructions is more than the length of the sequence the user expects to execute the same sequence multiple times

src/frontend/synthetic/kernel_factory.cc
```diff
@@ -0,0 +1,250 @@
...
+                                                                    uns num_of_dependence_chains) const {
+  switch (kernel) {
+    case MEM_BANDWIDTH_LIMITED:
+      return generate_load_kernel(NO_DEPENDENCE_CHAIN, workload_length, UNIFORM_SEQUENTIAL, (ICACHE_LINE_SIZE), 4,
+                                  DCACHE_LEVEL, get_start_pc(), get_start_uid());
+    case DCACHE_LIMITED:
+    case MLC_LIMITED:
+    case LLC_LIMITED:
+    case MEM_LIMITED:
+      return generate_load_kernel(DEPENDENCE_CHAIN, workload_length, UNIFORM_SEQUENTIAL, (2 * ICACHE_LINE_SIZE), 0,
```
hlitz: why icache line *2 ?

src/frontend/synthetic/kernel_factory.cc
```diff
@@ -0,0 +1,250 @@
...
+      num_of_ibr_ops_executed(0) {
+  /*
+    the ff are defaults, we edit them in the switch if we have to, for a particular workload.
+    If a particular workload does not require a field, its benign
+  */
+  std::cout << "Generating " << kernel_names[kernel] << " kernel" << std::endl;
+  Limit_Load_To level = DCACHE_LEVEL;
+  double t_nt_ratio = 1;
+  Sequence_Pick_Strategy branch_target_strategy = UNIFORM_SEQUENTIAL;
+  target_stride = 2 * ICACHE_LINE_SIZE;
```
hlitz: why *2 ?

src/frontend/synthetic/sampler.cc
hlitz: I don't understand the explanation. If the last address needs to point to the first, can we patch the last element's target instead? I think it's cleaner to have no if/else here

src/frontend/synthetic/synthetic_bottleneck_kernels.cc
```diff
@@ -0,0 +1,323 @@
...
+    // generate leading nops and return next pc
+    current_pc = gen_issue_width_lock_nops(kernel_map, uid_sequence, ISSUE_WIDTH - 1, current_pc);
+    auto current_uid = uid_sequence.get_next_element();
+
+    // every taken-target is 2 cachelines away.
+    _target = combined_targets_pool.peek_element_following_next();
+    next_inst = generate_conditional_branch(current_pc, current_uid, _target, direction_sequence.get_next_element(),
+                                            BRANCH_SIZE);
+
+    /* if the generated CBR's target goes out of range of target_pool, overwrite with unconditional branch to go back to
```
hlitz: can we ensure instead that this does not happen and replace it with an assert?

src/issue_queue.cc
hlitz: apperantly it is still part of the PR so it was not rebased correctly. This change should only exist in the other PR and not here

hlitz (changes requested):

src/frontend/synthetic/kernel_factory.cc
sweet12prof: I used the approach to ensure we exercise the bottleneck always, irrespective of the user-provided sequence length. 
If we want it to leave that semantic to the user, then I  guess the rerandomisation is unnecessary.

sweet12prof:
The following changes have been made: 

- Refactored the kernel factory into 2 classes; 
  - kernel_factory : Generates the kernel only
  - kernel_dispatcher : Instantiates a kernel, through a factory member, and is responsible for dispatching and cordinating execution with frontend. 
-  Refactored Kernel Generation Data structures 
   - Common data structures in the kernel APIs are defined as local members of kernel_factory. Extra APIs are defined in the Sampler class and exposed through public functions to dispatcher through the factory, so facilitate rerandomization in kernels like CBR. 
- Target Pools For CF workloads 
    - CBR kernel now uses a shuffled target pool, that is cyclic with every target repeating only once. The targets here remain fixed for the corresponding CBR pc over entire length of program.
    - Every CBR instruction is followed by an unconditional branch to the same target of the preceding CBR. This ensures fall-through paths do not create mutually branching pairs(infinite sub cycle in the execution graph).
    - UBR kernel also uses shuffled target pool with normal instructions branching to other targets other than the target of preceding UBRs. Mutually branching pairs are guaranteed to not exist here once the target pool is cyclic after shuffling. fall-through is offpath automatically. 
 
- Kernel Descriptions added in kernel header and, uniquely for each kernel definition.
    
- Issues. 
    The IBR workload needs some more thought.
    
 - In the Future
 For now the kernel definitions still remain in the synthetic_bottleneck.cc for consistence with where they were in previous commit. Will move them to kernel_factory.cc in the future commit where their review is over.

src/frontend/synthetic/synthetic_bottleneck_kernels.cc
```diff
@@ -0,0 +1,369 @@
...
+  for (uns i{0}; i < (target_pool_size); i++) {
+    // generate leading nops and return next pc
+    current_pc = target_pool.get_next_element();
+    current_pc = generate_nop_sequence(kernel_buffer, uid_sequence, DISPATCH_WIDTH, current_pc);
+    current_uid = uid_sequence.get_next_element();
+    // every taken target is the next element in the shuffled sequence
+    _target = target_pool.peek_next_element();
+    next_inst = generate_conditional_branch(current_pc, current_uid, _target, direction_sequence.get_next_element(),
+                                            get_inst_size());
+    kernel_buffer.push_back(next_inst);
```
hlitz: what is the purpose of the kernel buffer. It seems like extra overhead and intuitively also is confusing.

can't we just insert <PC,inst> into the map directly?

PR #547, opened 2026-07-12
Adding a perfect tag feature and more bp stats for alternative & aliasing issue

moonseoppkim:
Hi, @hlitz 

# SPEC2017 tage analysis

PCT

BASE PREDICTOR ACCURAY

SHORT TABLE ACCURACY

LONG TABLE ACCURAY

LOOP ACCURACY

SC ACCURACY

## ALTERNATIVE PREDICTOR PERFORMANCE (SHORT & LONG)

ALT_USAGE_PCT((SHORT_ALT+LONG_ALT) _ ALL_COMP)

ALT SHORT ACCURACY

ALT LONG ACCURACY

## ALIASING STAT ANALYSIS

ALIASING RATE

Aliasing rate: 
3.69% of predictions with an index & tag match had a stored-PC mismatch (0.5–7.5% per workload).

chart (13)

Mispredict rate, (aliased vs clean): 
If tag width were a problem, aliased hit should mispredict more than clean hit(no aliasing case). 
They don't: 2.6% (aliased) vs 4.0% (clean) on average. The only consistent exception is the gcc family (~2x worse), matching its large static branch footprint.

ALIAS_MISS _ (BASE_M+SHORT_M+LONG_M+LOOP_M+SC_M) vs  Workload

Alias share of all mispredictions: 
Out of every 100 mispredictions, at most ~3 mispredictions happen on an aliased hit (2.7% averaged across workloads). 

### Conclusion
Tag aliasing is not a meaningful misprediction source; tag width is not the bottleneck


hlitz (approved):
LGTM

PR #550, opened 2026-07-15
prefetcher: universal ChampSim shim (IPCP, MLOP, SPP, IPCP_FULL, IP-s…

yql5510718:
## What this is

A self-contained adapter that runs unmodified ChampSim (DPC-style) prefetchers inside Scarab's prefetch framework. Vendor prefetcher source is imported byte-for-byte; all Scarab awareness lives in the shim. Six prefetchers are included: IPCP, MLOP, SPP, a two-level cooperating IPCP, and the ChampSim baselines IP-stride and next-line.

## Structure & how an import works

    src/prefetcher/champsim/
    |-- csenv/            fake ChampSim environment + UNMODIFIED vendor .inc files
    |-- champsim_shim.*   shim core (address conversion, issue path, clock/stat bridges)
    |-- shim_import.h     import template: expands a vendor file into a full wrapper
    |-- wrappers/         one small .cc per import (state isolation requires one TU each)
    `-- pref_*.{param,stat}.def   per-prefetcher knobs/counters

Inbound (training): Scarab's existing per-level dispatchers (pref_dl0/umlc/ul1_{miss,hit,pref_hit}) call the import's pref_table.def hooks, gated by the instance lists from #543 -- the list a type appears in (--pref_mlc_prefetchers TYPE_IPCP,DEST_MLC) decides which cache level's traffic trains it. The wrapper strips Scarab's proc-id tag off the address, refreshes the fake warmup_complete[]/current_core_cycle[], and calls the vendor's own *_prefetcher_operate().

Outbound (emission): the vendor calls what it believes is ChampSim's CACHE::prefetch_line(); the fake csenv/cache.h forwards to shim_issue(), which re-stamps the proc-id (convert_to_cmp_addr) and enqueues through the framework's single routing primitive pref_addto_dest_req_queue() at the instance's DEST_* level. From that point the prefetch is indistinguishable from a native one (per-level queues, drain, MRT_DPRF, inclusive fills).

Feedback: demand pref-hits and L2 fills are bridged to note_pref_hit/note_pref_fill, so vendor-internal control loops (IPCP's per-class accuracy throttle) run on real Scarab events; MSHR occupancy is wired read-only so vendor backpressure heuristics (MLOP, IP-stride's adaptive fill level) see genuine state.

Adding a prefetcher = copy the vendor file verbatim + a stub:

    // wrappers/pref_nextline.cc -- complete file
    #define SHIM_NAME    nextline
    #define SHIM_UNAME   NEXTLINE
    #define SHIM_VENDOR  "prefetcher/champsim/csenv/next_line.l2c.inc"
    #define SHIM_API_L2C 1
    #include "prefetcher/champsim/shim_import.h"

plus a table row and a 3-line stat.def (~30 lines marginal cost). Optional knobs: SHIM_EXTRA_INIT/SHIM_EXTRA_OP (custom glue), SHIM_GEN_UL1_CACHE_FILL (eviction-trained vendors). One instance per shim type at one training level for now (asserted; the vendors keep state in file-scope globals, so multi-level instancing needs per-level stamping -- documented follow-up).

## Files changed (46, +5,199)

| category | LOC | role |
|---|---:|---|
| Vendor code | 3,414 | **unmodified** ChampSim/DPC source |
| Env fakes | 280 | glue |
| Shim core | 161 | glue |
| Import template | 199 | glue |
| Wrappers | 529 | glue |
| Decls + knobs | 229 | glue |
| Integration | 81 | table rows, build, stat/param regs |

**Vendor (never edited):** `csenv/{ipcp_isca2020.l1d, ipcp_isca2020.l2c, mlop_dpc3.l1d, spp.l2c, ip_stride.l2c, next_line.l2c}.inc`, `csenv/spp.h` — byte-for-byte copies.

**Env fakes:** `csenv/cache.h` (fake `CACHE`: `prefetch_line`→shim, MSHR/PQ occupancy, block state, pref-hit/fill feedback, metadata channel), `csenv/champsim.h` / `memory_class.h` / `ooo_cpu.h` (globals, `FILL_*`, types).

**Shim core:** `champsim_shim.{h,cc}` — proc-id tag strip/re-stamp, `shim_issue` → `pref_addto_dest_req_queue`, clock tick, stat bridge.

**Import template:** `shim_import.h` — expands registry-gated init, operate glue, and all nine hook trampolines from four `#define`s.

**Wrappers:** three template stubs (`nextline` 11 LOC, `spp` 13, `ipstride` 20) plus three hand-written where the glue is prefetcher-specific behavior: `ipcp` (per-class accuracy feedback, 149), `mlop` (fill routing + backpressure, 139), `ipcp_full_l1/l2` (L1→L2 metadata channel, 197).

**Decls + knobs:** `pref_champsim.h` (C-visible hook declarations), per-prefetcher `.param.def/.param.h/.stat.def`.

**Integration (only touches outside `champsim/`, all additive):** `pref_table.def` +6 rows, `stat_files.def`/`param_files.def` +11 lines, `CMakeLists.txt` +12, and a single include in `pref_common.c`.

### DIFF
--- a/src/prefetcher/champsim/csenv/cache.h
+++ b/src/prefetcher/champsim/csenv/cache.h
@@ -1,0 +1,188 @@
+/* Universal ChampSim shim: fake of ChampSim's inc/cache.h.
+ *
+ * This header is included by each VENDOR prefetcher source via its bare
+ * `#include "cache.h"`, which the compiler resolves to THIS file via
+ * current-file-directory lookup (the vendor .inc lives in this same csenv/ dir).
+ * Because the vendor source is wrapped in `namespace cs_<name> { ... }`, this
+ * CACHE class becomes `cs_<name>::CACHE` — a DISTINCT class per prefetcher, so
+ * `CACHE::l2c_prefetcher_operate` definitions in different vendor files never
+ * collide. (champsim.h / memory_class.h / champsim_shim.h are pre-included at
+ * GLOBAL scope by the wrapper, so their include guards make the includes below
+ * no-ops here — keeping NUM_CPUS, FILL_*, and ::champsim global.)
+ */
+#ifndef CHAMPSIM_COMPAT_CACHE_H
+#define CHAMPSIM_COMPAT_CACHE_H
+
+#include "champsim.h"
+#include "memory_class.h"
+#include "prefetcher/champsim/champsim_shim.h"
+
+/* Minimal per-block metadata a prefetcher may inspect (e.g. MLOP reads
+ * block[set][way].prefetch to detect a prefetch hit). */
+struct CS_BLOCK {
+  uint8_t prefetch;
+  uint8_t valid;
+};
+
+/* Minimal queue model for bandwidth-aware prefetchers (PQ.occupancy/SIZE,
+ * MSHR.occupancy/SIZE). Defaults make "queue never full" so prefetchers that
+ * gate on occupancy behave like the hand-ports (no PQ throttle). */
+struct CS_QUEUE {
+  uint32_t occupancy;
+  uint32_t SIZE;
+};
+
+class CACHE {
+ public:
+  /* ---- ChampSim-visible members vendor sources touch ---- */
+  uint32_t cpu;
+  const char* NAME;
+  uint32_t NUM_SET;
+  uint32_t NUM_WAY;
+  CS_QUEUE PQ;
+  CS_QUEUE MSHR;
+
+  /* ---- Scarab-side glue (not part of the ChampSim contract) ---- */
+  champsim::Level _level;   /* destination level (instance's DEST_* token) when !_route_by_fill */
+  bool _route_by_fill;      /* if true, route each prefetch by its fill_level arg
+                               (FILL_L1->DL0, FILL_L2->UMLC, FILL_LLC->UL1) so a
+                               multi-fill-level prefetcher hits all levels as designed */
+  bool _dedup_per_op;       /* if true, collapse duplicate blocks issued within one
+                               operate() to a single prefetch (e.g. MLOP re-issues a
+                               block when upgrading FILL_L2->FILL_L1; on a single
+                               Scarab level those are redundant). */
+  std::unordered_set<uint64_t> _op_issued;  /* blocks issued in the current operate */
+  /* Inter-level metadata channel: when set, every prefetch_line() this component
+   * issues also forwards (pf_addr, ip, metadata) downstream. Used to feed a
+   * cooperating next-level component (e.g. IPCP L1 -> L2), reproducing ChampSim's
+   * pf_metadata handoff where an L1 prefetch arriving at L2 triggers l2c operate. */
+  std::function<void(uint64_t /*pf_addr*/, uint64_t /*ip*/, uint32_t /*metadata*/)> _downstream_op;
+  int _hwp_id;
+  int _stat_issued;
+  int _stat_qfull;
+  /* Optional observability for the accuracy feedback loop (acc = useful/filled).
+   * Set by a wrapper to its prefetcher's stat ids; -1 = don't count. Lets us read
+   * the otherwise-internal pref_useful/pref_filled counters from memory.stat. */
+  int _stat_useful;
+  int _stat_fill;
+  uint8_t _cur_pref_hit;   /* set by wrapper before each operate; surfaced via block[][].prefetch */
+  CS_BLOCK _blk;           /* the single fake block get_set/get_way always resolve to */
+
+  /* Prefetch-accuracy accounting some prefetchers (IPCP-L1) read for degree
+   * adaptation. ChampSim's cache.cc updates these on fill/hit/MSHR-merge; the shim
+   * approximates that by attributing each successfully issued prefetch to a class
+   * (pf_metadata bits 8-11, ChampSim convention), then bumping pref_filled[class]
+   * at FILL (note_pref_fill, from the umlc_cache_fill hook) and pref_useful[class]
+   * on a demand hit to a prefetched line (note_pref_hit, from the pref-hit hooks).
+   * This restores IPCP's accuracy-based degree throttle; without it acc sticks at
+   * the 60% default and IPCP over-prefetches on irregular workloads. */
+  uint64_t pref_filled[NUM_CPUS][6] = {};
+  uint64_t pref_useful[NUM_CPUS][6] = {};
+  uint64_t pref_late[NUM_CPUS][6] = {};
+  uint64_t pf_fill = 0;
+  uint64_t pf_useful = 0;
+  /* block-addr -> class table so a later demand pref-hit can be attributed to the
+   * class that issued it (Scarab cache lines don't carry pf_metadata). Direct-mapped;
+   * collisions just misattribute a hit between classes (benign for the ratio). */
+  static const uint32_t CLASS_TBL_MASK = (1u << 16) - 1u;
+  uint8_t _pf_class_tbl[1u << 16] = {};
+  void note_pref_hit(uint64_t pf_addr) {
+    uint8_t cls = _pf_class_tbl[(pf_addr >> 6) & CLASS_TBL_MASK];
+    if (cls < 6)
+      pref_useful[cpu][cls]++;
+    champsim::shim_stat(cpu, _stat_useful);
+  }
+  /* A prefetch actually filled the cache: count it under its issuing class. acc =
+   * useful/filled then reflects true prefetch accuracy (fills that got demand-hit). */
+  void note_pref_fill(uint64_t fill_addr) {
+    uint8_t cls = _pf_class_tbl[(fill_addr >> 6) & CLASS_TBL_MASK];
+    if (cls < 6)
+      pref_filled[cpu][cls]++;
+    champsim::shim_stat(cpu, _stat_fill);
+  }
+
+  CACHE()
+      : cpu(0), NAME("champsim"), NUM_SET(1024), NUM_WAY(16),
+        PQ{0u, (1u << 20)}, MSHR{0u, (1u << 20)},
+        _level(champsim::Level::UMLC), _route_by_fill(false), _dedup_per_op(false),
+        _hwp_id(0), _stat_issued(-1), _stat_qfull(-1),
+        _stat_useful(-1), _stat_fill(-1), _cur_pref_hit(0) {
+    _blk.prefetch = 0;
+    _blk.valid = 1;
+    block.b = &_blk;
+  }
+
+  /* ---- ChampSim callbacks INTO the cache ---- */
+  int prefetch_line(uint64_t ip, uint64_t /*base_addr*/, uint64_t pf_addr,
+                    int fill_level, uint32_t metadata) {
+    if (_dedup_per_op) {
+      uint64_t blk = (pf_addr & ((1ULL << 58) - 1ULL)) >> 6;  /* LOG2_BLOCK_SIZE */
+      if (!_op_issued.insert(blk).second)
+        return 1;  /* already issued this operate; suppress redundant re-issue */
+    }
+    champsim::Level tgt = _route_by_fill ? champsim::level_from_fill(fill_level) : _level;
+    int ok = champsim::shim_issue(tgt, cpu, _hwp_id, pf_addr, _stat_issued, _stat_qfull);
+    if (ok) {
+      /* Remember block->class (from pf_metadata bits 8-11) so the eventual fill and
+       * demand pref-hit can be attributed to the issuing class. pref_filled is bumped
+       * at FILL (note_pref_fill), not here -- counting every issue overcounts (lookahead
+       * re-issues, drops, redundant hits), deflating acc and over-throttling. */
+      _pf_class_tbl[(pf_addr >> 6) & CLASS_TBL_MASK] = (uint8_t)((metadata >> 8) & 0x7u);
+    }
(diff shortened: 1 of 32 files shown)


hlitz (commented):
Some general questions/comments first:
- We generally have all params and stat in a single combined file. we can introduce 2 files for champsim but not one file for each prefetcher.
- does any of the champsim prefetchers distinguish between dcaache, l1, mlc ? ie does the code consider what level the prefetcher is installed at? If not why do we have 3 wrappers? one should be enough right?
I don't think the current structure makes a lot of sense. Either
a) Expose the individual champsim prefetchers to scarab. then we can use the existing per prefetcher function table to call the perf prefetcher APIs.
b) have all champsim prefetchers expose the exact same API name (overloaded from parent class) and then we don't need a wrapper.

It seems that champsim is just as idiotic as scarab and does not use a common overloaded API for all prefetchers so I assume a) would be a better way.

yql5510718:
Topsimpt compare against baseline(stream_ul1).
Workload | baseline IPC | mlop_ipcp_mlc IPC | % | spp_l1 IPC | %
-- | -- | -- | -- | -- | --
blender_r | 2.2213 | 2.2331 | +0.53% | 2.2387 | +0.78%
bwaves_r | 0.8638 | 0.8617 | -0.25% | 0.8279 | -4.16%
bwaves_r_2 | 0.8589 | 0.8571 | -0.22% | 0.8244 | -4.02%
bwaves_r_3 | 0.8499 | 0.8484 | -0.18% | 0.7681 | -9.63%
bwaves_r_4 | 0.8664 | 0.8665 | +0.01% | 0.8296 | -4.25%
cactuBSSN_r | 1.2993 | 1.5498 | +19.28% | 1.6528 | +27.21%
cam4_r | 1.3381 | 1.3454 | +0.55% | 1.3501 | +0.89%
fotonik3d_r | 1.0462 | 0.9323 | -10.88% | 1.0892 | +4.11%
imagick_r | 3.8285 | 3.9200 | +2.39% | 3.9073 | +2.06%
lbm_r | 1.1725 | 1.1197 | -4.51% | 0.9579 | -18.31%
nab_r | 1.3677 | 1.3819 | +1.04% | 1.3690 | +0.10%
namd_r | 3.5624 | 3.6836 | +3.40% | 3.4710 | -2.57%
parest_r | 1.6676 | 1.4598 | -12.46% | 1.5620 | -6.33%
povray_r | 3.0194 | 3.0196 | +0.01% | 3.0189 | -0.02%
roms_r | 1.3133 | 1.5951 | +21.45% | 1.7319 | +31.87%
wrf_r | 1.1772 | 1.1744 | -0.24% | 1.1525 | -2.09%
deepsjeng_r | 2.3622 | 2.3709 | +0.37% | 2.3609 | -0.06%
exchange2_r | 3.5953 | 3.5953 | 0.00% | 3.5953 | 0.00%
gcc_r | 1.5691 | 1.5735 | +0.28% | 1.5729 | +0.24%
gcc_r_2 | 1.7657 | 1.6735 | -5.22% | 1.7608 | -0.28%
gcc_r_3 | 1.1181 | 1.0319 | -7.71% | 1.3272 | +18.70%
gcc_r_4 | 1.9015 | 1.8558 | -2.40% | 1.9164 | +0.78%
gcc_r_5 | 1.8911 | 1.9066 | +0.82% | 1.9112 | +1.06%
leela_r | 1.6315 | 1.6326 | +0.07% | 1.6321 | +0.04%
mcf_r | 0.7713 | 0.7095 | -8.01% | 0.7212 | -6.49%
omnetpp_r | 0.9308 | 0.9482 | +1.86% | 0.9563 | +2.74%
perlbench_r | 2.7477 | 2.8391 | +3.33% | 2.8118 | +2.34%
perlbench_r_2 | 3.1037 | 3.1094 | +0.18% | 3.1070 | +0.11%
perlbench_r_3 | 2.0992 | 2.3348 | +11.22% | 2.2057 | +5.07%
x264_r | 4.0981 | 4.2604 | +3.96% | 4.1827 | +2.06%
x264_r_2 | 4.4285 | 4.5444 | +2.62% | 4.5124 | +1.89%
x264_r_3 | 4.3331 | 4.4802 | +3.39% | 4.4070 | +1.71%
xalancbmk_r | 1.0500 | 1.1803 | +12.41% | 1.1153 | +6.21%
xz_r | 0.8252 | 0.8178 | -0.89% | 0.8207 | -0.54%
xz_r_2 | 2.2965 | 2.2924 | -0.18% | 2.3085 | +0.52%
xz_r_3 | 1.5858 | 1.5663 | -1.23% | 1.5867 | +0.05%
Avg | 1.9599 | 1.9881 | +1.44% | 1.9879 | +1.43%

src/prefetcher/champsim/wrappers/pref_mlop.cc
```diff
@@ -0,0 +1,141 @@
...
+  STAT_EVENT(proc_id, PREF_SHIM_OPERATE);
+  champsim::tick(proc_id);
+  g_mlop.cpu = proc_id;
+  g_mlop.begin_operate(); /* reset per-operate dedup set */
+  /* Real backpressure (READ-ONLY Scarab state) so MLOP's built-in PQ/MSHR throttle
+   * self-limits like ChampSim (MEM_REQ_BUFFER_ENTRIES==32 matches ChampSim MSHR). */
+  g_mlop.MSHR.SIZE = MEM_REQ_BUFFER_ENTRIES;
+  g_mlop.MSHR.occupancy = mem_get_req_count(proc_id);
+  g_mlop.PQ.SIZE = PREF_UMLC_REQ_QUEUE_SIZE;
+  g_mlop.PQ.occupancy = mem->uncores[proc_id].num_outstanding_l1_misses;
```
hlitz: `PQ.SIZE` is hardcoded to the UMLC queue, but MLOP can be placed at any single level — `_level` comes from `pref_hwp_instance_dest(hwp, lvl)` where `lvl` is DCACHE, UMLC, or UL1. When TYPE_MLOP is listed in `--pref_dcache_prefetchers` or `--pref_l1_prefetchers`, the vendor's built-in PQ throttle (`mlop_dpc3.l1d.inc:652`) is gated on the wrong level's capacity.

Suggest deriving it from the instance's dest level, e.g. a small `shim_queue_size(champsim::Level)` in `champsim_shim.cc` returning `PREF_DL0_REQ_QUEUE_SIZE` / `PREF_UMLC_REQ_QUEUE_SIZE` / `PREF_UL1_REQ_QUEUE_SIZE`, so this reads `g_mlop.PQ.SIZE = champsim::shim_queue_size(g_mlop._level);`.

Related: `PQ.occupancy = num_outstanding_l1_misses` is a proxy for a different quantity than the dest prefetch queue's occupancy. If the framework can expose the real queue fill for the level, that would be the faithful input; if not, worth a comment saying it's an approximation.

src/prefetcher/champsim/csenv/cache.h
```diff
@@ -0,0 +1,188 @@
...
+  void note_pref_fill(uint64_t fill_addr) {
+    uint8_t cls = _pf_class_tbl[(fill_addr >> 6) & CLASS_TBL_MASK];
+    if (cls < 6)
+      pref_filled[cpu][cls]++;
+    champsim::shim_stat(cpu, _stat_fill);
+  }
+
+  CACHE()
+      : cpu(0), NAME("champsim"), NUM_SET(1024), NUM_WAY(16),
+        PQ{0u, (1u << 20)}, MSHR{0u, (1u << 20)},
```
hlitz: The ctor defaults `PQ.SIZE`/`MSHR.SIZE` to `1<<20`, i.e. "queue never full". Only `pref_mlop` and `pref_ipstride` override them; `pref_ipcp`, `pref_spp`, and `pref_nextline` do not. Any vendor prefetcher that self-throttles on occupancy therefore runs with its throttle disabled, and it fails *silently* — no build error, no assert, just a prefetcher that is more aggressive than the paper it came from.

Two options, either is fine:
1. Wire the real values in the generic path — set `PQ`/`MSHR` from Scarab state inside `begin_operate()` (or in `shim_import.h`'s `shim_op`, which already runs per operate) so every import gets backpressure by default and `SHIM_EXTRA_OP` is only for extras.
2. Keep the override model but make the default fail loudly: initialize to 0 and assert on first `operate()` that a wrapper set them.

Option 1 also removes the duplicated MSHR wiring currently copy-pasted into `pref_ipstride.cc:17-18` and `pref_mlop.cc:99-100`.

src/prefetcher/champsim/wrappers/pref_mlop.cc
```diff
@@ -0,0 +1,141 @@
...
+  /* Default: collapse all fills onto the drive level (best IPC in Scarab's tiny
+   * L1D). Set --pref_mlop_route_by_fill 1 for faithful multi-fill (FILL_L1->L1D,
+   * FILL_L2->L2; needs --pref_dl0_fill_dcache for real L1D fills). */
+  g_mlop._route_by_fill = PREF_MLOP_ROUTE_BY_FILL;
+  g_mlop._hwp_id = hwp->hwp_info->id;
+  g_mlop._stat_issued = PREF_SHIM_ISSUED;
+  g_mlop._stat_qfull = PREF_SHIM_QFULL;
+  /* Cache geometry MLOP uses only to size its access-map table. */
+  g_mlop.NUM_SET = 1024;
+  g_mlop.NUM_WAY = 16;
```
hlitz: These two magic numbers deserve a sharper comment than "geometry MLOP uses only to size its access-map table", because they are not geometry — they are a storage budget, and they do not correspond to any cache in this simulator.

1024x16 = 16384 blocks implies a 1MB L1D. Scarab's defaults are `dcache_size` 64KB / `dcache_assoc` 1 (`memory/memory.param.def:145-146`) and `mlc_size` 512KB / `mlc_assoc` 4. Via `AMT_SIZE = 32 * NUM_SET * NUM_WAY / BLOCKS_IN_ZONE` (`mlop_dpc3.l1d.inc:825-827`) this gives MLOP an 8192-entry access-map table; tying it to the real dest level's geometry would shrink it ~16x and change results.

So the value is a deliberate choice, not a bug — but it should be explicit about which it is: either derive it from the dest level's real `*_SIZE`/`*_ASSOC` params, or keep the fixed numbers and say in the comment that they are ChampSim-parity table-budget constants intentionally decoupled from Scarab's cache config (ideally with a param so a sweep can vary the AMT budget).

hlitz (commented):
Three notes on the fake `CACHE` in `csenv/cache.h` and how the wrappers feed it. The class itself is the right call (vendor sources are written as `CACHE::` methods, and the per-namespace trick is what avoids the ODR collision across five imports) — these are all about the handful of fields that feed real prefetcher *decisions*, where the shim silently diverges from Scarab state.

src/prefetcher/champsim/csenv/champsim.h
```diff
@@ -0,0 +1,57 @@
...
+ * NUM_CORES <= NUM_CPUS. */
+#define NUM_CPUS 1
+
+#define PAGE_SIZE 4096
+#define LOG2_PAGE_SIZE 12
+#define BLOCK_SIZE 64
+#define LOG2_BLOCK_SIZE 6
+
+/* ChampSim cache sizing constants some prefetchers use for table sizing. */
+#define L2C_MSHR_SIZE 32
```
hlitz: This should be renamed, and the comment above it changed. Despite the name, the only consumer does not use it as an MSHR model at all -- it is the extent of SPP's lookahead queue:

```c
uint32_t confidence_q[L2C_MSHR_SIZE], ...   // spp.l2c.inc:69
int32_t  delta_q[L2C_MSHR_SIZE];            // spp.l2c.inc:73
```

and `PATTERN_TABLE::read_pattern` writes into them with no bounds check -- `confidence_q[pf_q_tail] = pf_conf; ... pf_q_tail++;` (spp.l2c.inc:503-511), inside a `do { ... } while (do_lookahead)` loop that can append up to `PT_WAY` entries per iteration. Nothing compares `pf_q_tail` to `L2C_MSHR_SIZE`; 32 is empirical headroom that works because lookahead confidence decays below `PF_THRESHOLD` first.

So the name actively invites a change that would be a stack buffer overflow: `#define L2C_MSHR_SIZE MEM_REQ_BUFFER_ENTRIES` (a runtime param, `memory/memory.param.def:168`) would (a) make those two declarations VLAs, and (b) let `--mem_req_buffer_entries 8` corrupt the stack on any deep lookahead. The heading comment ("cache sizing constants some prefetchers use for table sizing") does not warn about either.

The vendor spelling has to stay, since `spp.l2c.inc` is byte-identical -- so suggest renaming the authoritative definition and aliasing:

```c
/* SPP's lookahead-queue depth: the extent of spp.l2c.inc's confidence_q/delta_q.
 * NOT an MSHR model, and must stay a compile-time constant -- read_pattern()
 * advances pf_q_tail unchecked, so do NOT derive this from
 * MEM_REQ_BUFFER_ENTRIES or any other runtime param. 32 = ChampSim/SPP parity. */
#define CHAMPSIM_LOOKAHEAD_Q_SIZE 32
#define L2C_MSHR_SIZE CHAMPSIM_LOOKAHEAD_Q_SIZE  /* vendor-facing name; keep as-is */
```

Worth contrasting explicitly with `CACHE::MSHR.SIZE`, which *is* live Scarab state and correctly tracks `MEM_REQ_BUFFER_ENTRIES` (`pref_ipstride.cc:17`) because it is only ever read in an occupancy comparison. Two things named MSHR, opposite rules: one frozen array bound, one live simulator value.

PR #552, opened 2026-07-25
LVP: in-place off-path recovery for mispredicted loads (WIP)

hlitz:
## Summary

Draft PR for review of the **load-value/address misprediction recovery** mechanism on top of the generalized LVP framework (branch `lvp-on-main`). Base is `main`, so the diff also includes the two prior LVP-framework commits (`Load Value Prediction ported onto main`, `Generalize load predictor into value + address framework`); the new work is in the top commit `LVP: in-place off-path recovery for mispredicted loads`.

## What this does

**Frontend (decoupled) — in-place off-path redirect (no FT split):**
- New `FT_EVENT_LOAD_MISPREDICT`. On a mispredicted on-path load, `redirect_load_mispred()` moves the on-path consumers after the load's EOM into a saved recovery FT and refills the *same* FT's tail with off-path ops (no split, no deep copy).
- The load's EOM is marked `op->load_pred_offpath_after`; `icache_process_ops` flips `ic->off_path` right after fetching it — mirroring the branch `recover_at_*` flip — so the `ic->off_path == op->off_path` invariant holds across the on→off boundary inside one mixed FT.

**Rename / register file (produce-at-rename):**
- Predicted loads produce at rename (`map_stage`) via `wake_up_ops`, with `produced_cycle = wake_cycle`, so early-woken consumers don't violate `onpath_consumed_cycle >= produced_cycle`.
- `produced_cycle` bookkeeping in `map_rename` (produce-once assert; retire-time tripwire).

**dcache / exec:**
- Suppress the duplicate produce for predicted loads.
- Guard against double-scheduling recovery (`recover_at_exec && !recovery_sch`).

## Status / known issue (WIP)

The current tip **still crashes** at `map_rename.c:1059` (`ASSERT(op->inst_info->table_info.cf_type)`): the rename recovery path assumes the recovery point is a branch, and `reg_file_snapshot_srt` is only taken for branches. The fix under discussion:
1. Snapshot the SRT at rename for load recovery points too, keyed on `load_pred_offpath_after` (a **fetch-time** marker — required because value-mode `recover_at_exec` isn't set until load completion, after rename).
2. Relax the `cf_type` assert in `reg_renaming_scheme_realistic_recover`.

Not yet applied — deliberately left out so the recovery approach can be reviewed first.

## Validation so far
- Builds clean.
- `baseline` config: passes end-to-end on the SPEC2017 `rate_int_v2` exec sweep.
- Prediction configs (`value_last_value`, `addr_const_rfp`): clear the earlier `icache_stage.c:897` mixed-FT assert (the icache off-path flip works); now stop at the `map_rename.c:1059` recovery assert above.

## Review asks
- Is the in-place off-path redirect (vs. FT split) the right approach?
- The `load_pred_offpath_after` marker now serves double duty (icache flip + SRT snapshot trigger) — rename to `load_pred_recovery_point`?
- The SRT-snapshot-at-rename keyed on a fetch-time marker (single-checkpoint model): sound?


src/exec_stage.c
```diff
@@ -546,15 +546,16 @@ static inline void exec_stage_process_op(Op* op) {
   STAT_EVENT(op->proc_id, EXEC_ON_PATH_INST_MEM + (op->inst_info->table_info.mem_type == NOT_MEM) + 2 * op->off_path);
   STAT_EVENT(op->proc_id, EXEC_ALL_INST);
 
-  STAT_EVENT(op->proc_id, LOAD_VALUE_PREDICT_SAVED_CYCLES_ON_PATH);
+  if (op->load_value_predicted && !op->off_path)
+    STAT_EVENT(op->proc_id, LOAD_VALUE_PREDICT_SAVED_CYCLES_ON_PATH);
```
hlitz: does this really count saved cycles due to vp? It seems to only count the number of on-path value predicted ops.

src/exec_stage.c
hlitz: remove load_value_flush here and add a new path to dcache where if if load completes and the load is value predicted, check if value _flush is set and then schedule recovery at done_cycle

src/exec_stage.c
```diff
@@ -566,6 +567,8 @@ static inline void exec_stage_bp_resolve(Op* op) {
   if (op->bp_pred_info->recover_at_exec) {
     DEBUG(exec->proc_id, "Exec schedules recovery for op_num:%llu at cycle:%llu\n", (unsigned long long)op->op_num,
           (unsigned long long)op->exec_cycle);
+    // load value/addr flush is not a branch: skip branch-resolve latency stat
```
hlitz: remove all load_val_related stuff as we no longer call bp_resolve for value pred loads

src/ft.cc
```diff
@@ -398,18 +398,23 @@ FT_PredictResult FT::predict_ft() {
...
-          event = FT_EVENT_MISPREDICT;
-        }
+    // Load value/address prediction (main BP pass only, so it runs once per op).
+    // Predict/resolve a load early; on a wrong prediction, stamp the squash on
+    // this macro's EOM op. This is data (not control) speculation: the fetched
+    // path does not change, so we do NOT redirect the frontend here. The op is
+    // only marked recover_at_exec - the redirect/refetch happens after recovery
+    // fires at exec (bp_sched_recovery). Stamping the EOM (not a mid-macro uop)
+    // avoids skipping siblings; the EOM is guaranteed present (an FT ends on eom).
+    if (bp_id == MAIN_BP && !op->inst_info->table_info.cf_type) {
```
hlitz: instead of no cf check for mem_type

src/icache_stage.c
```diff
@@ -980,10 +980,6 @@ static inline void icache_process_ops(Stage_Data* cur_data, Flag fetched_from_uo
       op->bp_pred_l0.pred_global_hist = g_bp_data->global_hist;
       op->bp_pred_main.pred_global_hist = g_bp_data->global_hist;
     }
-
-    if (op->load_value_flush) {
```
hlitz: why is this beeig removed? don't we need to put ic offpath as soon as we see a value_flush

src/load_value_pred.cc
```diff
@@ -43,102 +53,87 @@ extern "C" {
...
-  for (uns ii = 0; ii < inst_info->table_info.num_src_regs; ii++)
-    printf("%d, ", inst_info->srcs[ii].id);
-  printf(">, dest#%d: <", inst_info->table_info.num_dest_regs);
-  for (uns ii = 0; ii < inst_info->table_info.num_dest_regs; ii++)
-    printf("%d, ", inst_info->dests[ii].id);
-  printf(">\n");
+static inline Flag load_pred_read_dest_value(Op* op, uns64* value) {
+  if (op->inst_info->table_info.num_dest_regs < 1)
+    return FALSE;
+  *value = op->dst_val[0];
```
hlitz: we need to read the correct corresponding dst_val of a given ID instead of reading 0 always

src/load_value_pred.cc
```diff
@@ -43,102 +53,87 @@ extern "C" {
...
   op->load_value_flush = TRUE;
 
-  const Addr pc_plus_offset = ADDR_PLUS_OFFSET(op->inst_info->addr, op->inst_info->trace_info.inst_size);
-  op->oracle_info.dir = NOT_TAKEN;
-
-  // Branch-prediction/recovery state moved to bp_pred_info on modern main; set
-  // both levels so the exec-stage recovery check fires regardless of selection.
+  // On-path non-CF ops select bp_pred_main during predict_ft; set both levels so
+  // the exec-stage recovery check fires regardless of which is selected.
   for (Bp_Pred_Info* bp : {&op->bp_pred_main, &op->bp_pred_l0}) {
```
hlitz: only set main, not l0

src/load_value_pred.cc
```diff
@@ -43,102 +53,87 @@ extern "C" {
...
+ * on this macro's EOM op (it owns the FT vector and can look forward to the EOM),
+ * so no cross-op global latch is needed.
+ */
+static inline void load_pred_apply_early_resolve(Op* op, Flag is_mispred) {
   ASSERT(op->proc_id, op->inst_info->table_info.mem_type == MEM_LD);
 
-  // set the metadata to indicate bypassing
   op->load_value_predicted = TRUE;
   op->decode_cycle = cycle_count;
   op->exec_cycle = cycle_count;
```
hlitz: don't we need to call wake-up here and write to regfile or is this done elsehwere?

src/load_value_pred.cc
```diff
@@ -171,25 +168,107 @@ class LoadValuePredictor {
...
+    ASSERT(op->proc_id, op->inst_info->table_info.mem_type == MEM_LD);
+    auto* pred_entry = static_cast<LastValuePredEntry*>(entry);
+    if (!pred_entry->is_found())
+      return;
+    if (pred_entry->confidence <= LOAD_VALUE_PRED_THRESHOLD)
+      return;
+
+    // SEAM: requires the load's true destination value to verify the prediction.
+    uns64 actual_value = 0;
+    if (!load_pred_read_dest_value(op, &actual_value))
```
hlitz: need to iterate over valid_dests and predict each one. either here or maybe call infer and train for all dests so we do not need to replicate code in the overloaded train/infer functions

src/load_value_pred.cc
```diff
@@ -171,25 +168,107 @@ class LoadValuePredictor {
...
+  }
+
+  void train(Op* op, PredictorEntry* entry) override {
+    ASSERT(op->proc_id, op->inst_info->table_info.mem_type == MEM_LD);
+    if (op->off_path)
+      return;
+
+    // SEAM: requires the load's true destination value to update the table.
+    uns64 actual_value = 0;
+    if (!load_pred_read_dest_value(op, &actual_value))
```
hlitz: iterate over dests

src/ft.cc
```diff
@@ -399,22 +399,45 @@ FT_PredictResult FT::predict_ft() {
...
           eom_idx++;
         ASSERT(proc_id, eom_idx < ops.size());
         // the EOM must belong to the same macro-instruction as the load
         ASSERT(proc_id, ops[eom_idx]->inst_uid == op->inst_uid);
-        load_pred_schedule_squash(ops[eom_idx]);
+        Op* eom = ops[eom_idx];
+        // Record the recovery trigger: AGEN-verified fires at the EOM's exec
+        // (recover_at_exec); value-verified fires at load completion (dcache,
+        // load_pred_verify_at_completion) -- record the EOM for it.
+        if (!op->load_pred_verify_at_done)
```
hlitz: why is this if/else case needed. we should be able to just always set recovery_at_exec for the eom and thats it for here

src/op.h
```diff
@@ -202,6 +202,13 @@ struct Op_struct {
   Flag load_value_predicted;      // if consumers of the op can be ready before this load
   Flag load_value_flush;          // if this is the eom of a load value/addr predict inst
   Flag load_value_mispredicted;   // load's value/addr was mispredicted (EOM will schedule squash)
+  Flag load_addr_predicted;       // early-AGEN: load may issue w/o addr operands, access load_pred_addr
+  Addr load_pred_addr;            // predicted effective addr (early-AGEN/RFP), verified vs va at exec
+  Counter load_pred_ready_cycle;  // cycle a predicted load's result is available to consumers (RFP: +DCACHE_CYCLES)
+  Flag load_pred_verify_at_done;  // value pred: verify + squash at load completion (done), not at AGEN
```
hlitz: remove these fields and instead introduce recover_at_agen and recover_at_load_completion for bp_recovery info. in FT where the mispred is detected set the correct recovery flag based on the reason.

src/op.h
```diff
@@ -202,6 +202,13 @@ struct Op_struct {
...
+  struct Op_struct* load_pred_squash_op;  // macro EOM the deferred value-pred squash targets (recorded at FT)
+  Counter load_pred_squash_unique;        // load_pred_squash_op's unique_num, validated at completion
+  Flag load_pred_offpath_after;   // load-mispred EOM: icache flips off-path AFTER fetching this op (mixed FT)
```
hlitz: also seems redundant. we can just check in icache for value_mispred and then go ooffpath

src/op.h
```diff
@@ -202,6 +202,13 @@ struct Op_struct {
   Flag load_value_predicted;      // if consumers of the op can be ready before this load
   Flag load_value_flush;          // if this is the eom of a load value/addr predict inst
   Flag load_value_mispredicted;   // load's value/addr was mispredicted (EOM will schedule squash)
+  Flag load_addr_predicted;       // early-AGEN: load may issue w/o addr operands, access load_pred_addr
+  Addr load_pred_addr;            // predicted effective addr (early-AGEN/RFP), verified vs va at exec
+  Counter load_pred_ready_cycle;  // cycle a predicted load's result is available to consumers (RFP: +DCACHE_CYCLES)
+  Flag load_pred_verify_at_done;  // value pred: verify + squash at load completion (done), not at AGEN
+  struct Op_struct* load_pred_squash_op;  // macro EOM the deferred value-pred squash targets (recorded at FT)
+  Counter load_pred_squash_unique;        // load_pred_squash_op's unique_num, validated at completion
```
hlitz: why is this field needed

src/op.h
```diff
@@ -202,6 +202,13 @@ struct Op_struct {
   Flag load_value_predicted;      // if consumers of the op can be ready before this load
   Flag load_value_flush;          // if this is the eom of a load value/addr predict inst
   Flag load_value_mispredicted;   // load's value/addr was mispredicted (EOM will schedule squash)
+  Flag load_addr_predicted;       // early-AGEN: load may issue w/o addr operands, access load_pred_addr
+  Addr load_pred_addr;            // predicted effective addr (early-AGEN/RFP), verified vs va at exec
+  Counter load_pred_ready_cycle;  // cycle a predicted load's result is available to consumers (RFP: +DCACHE_CYCLES)
+  Flag load_pred_verify_at_done;  // value pred: verify + squash at load completion (done), not at AGEN
+  struct Op_struct* load_pred_squash_op;  // macro EOM the deferred value-pred squash targets (recorded at FT)
```
hlitz: why is this field needed

src/load_value_pred.cc
```diff
@@ -109,18 +111,53 @@ void load_pred_schedule_squash(Op* op) {
...
+
+  // Stamp the squash on the macro's EOM (recorded at FT in ft.cc), not on this
+  // possibly mid-macro load uop. For single-uop loads the EOM is the load
+  // itself. The EOM is a younger sibling on the same path, so it is still live
+  // (not yet retired) at the load's completion.
+  Op* eom = op->load_pred_squash_op;
+  ASSERT(op->proc_id, eom && eom->op_pool_valid && eom->unique_num == op->load_pred_squash_unique);
+
+  load_pred_schedule_squash(eom);
+  bp_sched_recovery(bp_recovery_info, eom, op->done_cycle);
```
hlitz: i think bp_sched_recovery should be in pred_schedule_squash

src/load_value_pred.cc
```diff
@@ -109,18 +111,53 @@ void load_pred_schedule_squash(Op* op) {
...
+  eom->recovery_scheduled = TRUE;
```
hlitz, line 128: why is this field needed if it is move in schedule_squash

src/load_value_pred.cc
```diff
@@ -109,18 +111,53 @@ void load_pred_schedule_squash(Op* op) {
...
+ * Value-prediction verification at load completion.  Called from the dcache stage
+ * when a load finishes (its real value is now known).  A value predictor verifies
+ * at completion rather than at AGEN, so if the predicted value was wrong the
+ * squash is scheduled here, at the load's done_cycle - modelling that the
+ * misprediction is only detected once the load returns its data.  The recovery is
+ * stamped on the load itself and fires immediately (bp_sched_recovery), flushing
+ * the speculatively-woken consumers so they re-execute with the correct value.
  */
-static inline void load_pred_apply_early_resolve(Op* op, Flag is_mispred) {
+void load_pred_verify_at_completion(Op* op) {
```
hlitz: rename into predicted_load_schedule_recovery

src/load_value_pred.cc
```diff
@@ -109,18 +111,53 @@ void load_pred_schedule_squash(Op* op) {
...
+  if (!op->load_value_mispredicted || !op->load_pred_verify_at_done || op->off_path)
```
hlitz: checks can be removed

src/dcache_stage.c
```diff
@@ -277,10 +285,13 @@ void update_dcache_stage(Stage_Data* src_sd) {
...
 
       op->done_cycle = cycle_count + DCACHE_CYCLES + op->inst_info->extra_ld_latency;
-      if (op->inst_info->table_info.mem_type != MEM_ST) {
+      // A value-predicted / RFP load already produced its register and woke its
+      // consumers at rename; do not wake (and re-produce) again here.
+      if (op->inst_info->table_info.mem_type != MEM_ST && !op->load_value_predicted) {
         op->wake_cycle = op->done_cycle;
         wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
       }
+      load_pred_verify_at_completion(op);
```
hlitz: why only for PERFECT_DCACHE?

take this out and just do: if (op->bp->recover_at_agen) predicted_load_schedule_recov()

src/dcache_stage.c
```diff
@@ -578,11 +589,13 @@ static inline void dcache_cacheline_hit(Op* op, Addr line_addr, Dcache_Data* lin
...
 
-  /* wake up source inst if the op is completed */
-  if (op->inst_info->table_info.mem_type != MEM_ST) {
+  /* wake up source inst if the op is completed (skip if already produced early
+   * by value/RFP prediction at rename) */
+  if (op->inst_info->table_info.mem_type != MEM_ST && !op->load_value_predicted) {
     op->wake_cycle = op->done_cycle;
     wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
   }
+  load_pred_verify_at_completion(op);
```
hlitz: take this out and just do: if (op->bp->recover_at_load_completion) predicted_load_schedule_recov()

src/dcache_stage.c
```diff
@@ -608,8 +621,12 @@ static inline void dcache_cacheline_miss(Op* op, Addr line_addr) {
...
 
         op->done_cycle = cycle_count + DCACHE_CYCLES + op->inst_info->extra_ld_latency;
-        op->wake_cycle = cycle_count + DCACHE_CYCLES + op->inst_info->extra_ld_latency;
-        wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
+        // skip re-wake if produced early by value/RFP prediction at rename
+        if (!op->load_value_predicted) {
+          op->wake_cycle = cycle_count + DCACHE_CYCLES + op->inst_info->extra_ld_latency;
+          wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
+        }
+        load_pred_verify_at_completion(op);
```
hlitz: same here on miss it is not rcompleted right?

src/dcache_stage.c
```diff
@@ -860,10 +877,12 @@ static inline void dcache_fill_process_cacheline(Mem_Req* req, Dcache_Data* data
     op->done_cycle = cycle_count + 1;
     op->state = OS_SCHEDULED;
 
-    if (op->inst_info->table_info.mem_type != MEM_ST) {
+    // skip re-wake if produced early by value/RFP prediction at rename
+    if (op->inst_info->table_info.mem_type != MEM_ST && !op->load_value_predicted) {
       op->wake_cycle = op->done_cycle;
       wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
     }
+    load_pred_verify_at_completion(op);
```
hlitz: take this out and just do: if (op->bp->recover_at_load_completion) predicted_load_schedule_recov()

src/icache_stage.c
```diff
@@ -981,12 +981,12 @@ static inline void icache_process_ops(Stage_Data* cur_data, Flag fetched_from_uo
...
-    // on-path op of a mixed FT whose tail was refilled off-path in place by the
-    // decoupled frontend. Flip the icache off-path here, mirroring the branch
-    // flip above (recover_at_*), so the following tail ops satisfy the
+    // Load value/address mispredict: the decoupled frontend refilled this FT's
+    // tail off-path in place after the mispredicted load. Flip the icache
+    // off-path right after the load, mirroring the branch flip above
+    // (recover_at_*), so the following tail ops satisfy the
     // ic->off_path == op->off_path invariant.
-    if (op->load_pred_offpath_after) {
+    if (op->load_value_mispredicted && !op->off_path) {
```
hlitz: is this the only sue of load_value_mispredicted? we can eliminate that field if we just check recover_at_agen or recover_at_load_completion

src/ft.cc
```diff
@@ -400,43 +400,24 @@ FT_PredictResult FT::predict_ft() {
...
-        while (eom_idx < ops.size() && !ops[eom_idx]->eom)
-          eom_idx++;
-        ASSERT(proc_id, eom_idx < ops.size());
-        // the EOM must belong to the same macro-instruction as the load
-        ASSERT(proc_id, ops[eom_idx]->inst_uid == op->inst_uid);
-        Op* eom = ops[eom_idx];
-        // Record the recovery trigger: AGEN-verified fires at the EOM's exec
-        // (recover_at_exec); value-verified fires at load completion (dcache,
-        // load_pred_verify_at_completion) -- record the EOM for it.
-        if (!op->load_pred_verify_at_done)
```
hlitz: where are we assigning recover_at_X now?

src/bp/bp.c
```diff
@@ -1005,6 +1005,12 @@ void bp_retire_op(Bp_Data* bp_data, Op* op) {
 void bp_recover_op(Bp_Data* bp_data, Cf_Type cf_type, Recovery_Info* info) {
   STAT_EVENT(0, PERFORMED_RECOVERIES);
   INC_STAT_EVENT(0, PERFORMED_RECOVERY_LAT, cycle_count - info->predict_cycle);
+
+  /* A load value/addr flush rides the recovery path as a non-CF op: it must not
```
hlitz: remove this and guard against calling bp_recover_op in the first place

src/load_value_pred.cc
```diff
@@ -0,0 +1,606 @@
...
+ * On-path non-CF loads select bp_pred_main during predict_ft, so the flag lives
+ * there.  The reason is set at predict time; only one is ever set per load.
+ */
+static inline void load_pred_mark_recovery(Op* op, Flag recover_at_completion) {
+  op->load_value_mispredicted = TRUE;
+  if (recover_at_completion)
+    op->bp_pred_main.recover_at_load_completion = TRUE;
+  else
+    op->bp_pred_main.recover_at_agen = TRUE;
+  if (!op->off_path)
```
hlitz: remove If and instead add an assertion. we should never schedule recovery if we are off-path already

src/load_value_pred.cc
```diff
@@ -0,0 +1,606 @@
...
+
+/*
+ * Schedule the recovery for a mispredicted predicted load.  Called from the
+ * dcache stage, gated by the load's recover_at_agen / recover_at_load_completion
+ * flag, so the reason and timing are already decided; this only needs to fire the
+ * squash once.  Clearing load_value_mispredicted makes a replayed load idempotent
+ * (bp_sched_recovery asserts a recovery is not scheduled twice).
+ */
+void predicted_load_schedule_recovery(Op* op, Counter recover_cycle) {
+  if (!op->load_value_mispredicted)
```
hlitz: no we should never recover twice, if we do its a bug. remove the if and replace with assertion that load_value_mispred is true

src/dcache_stage.c
```diff
@@ -253,10 +261,16 @@ void update_dcache_stage(Stage_Data* src_sd) {
...
     Addr line_addr;
-    Dcache_Data* line = (Dcache_Data*)cache_access(&dc->dcache, op->oracle_info.va, &line_addr, TRUE);
+    Dcache_Data* line = (Dcache_Data*)cache_access(&dc->dcache, access_va, &line_addr, TRUE);
     op->dcache_cycle = cycle_count;
     dc->idle_cycle = MAX2(dc->idle_cycle, cycle_count + DCACHE_CYCLES);
 
+    // Address prediction (early-AGEN / RFP) verifies at AGEN: the effective
+    // address is known once the load accesses the dcache, so a mispredicted
+    // predicted address schedules its recovery here, on hit or miss alike.
+    if (op->bp_pred_info->recover_at_agen)
```
hlitz: 1. recover_at_agen needs to be enabled for the load-address_generating uop. (an ALU op)
2. move all recover_at_agen/exec/load_complete into set_exec_cycle()
3. we may not even need load_complete/agen and can use recover_at_exec for all ops

src/dcache_stage.c
```diff
@@ -245,16 +253,6 @@ void update_dcache_stage(Stage_Data* src_sd) {
       STAT_EVENT(dc->proc_id, DCACHE_READ_PORT_UNAVAILABLE_ONPATH + op->off_path);
       continue;
     }
-    // Record the first dcache access, keeping dcache_cycle write-once. The only
-    // legitimate re-probe is a miss still waiting for a mem-request buffer
-    // (OS_WAIT_MEM): it stays resident and re-runs this access every cycle. Checked
-    // here, before op->state is set OS_SCHEDULED below, so op->state is still the
-    // entry state; any re-probe from a non-WAIT_MEM state is a bug.
-    if (op_get_dcache_cycle(op) == MAX_CTR)
```
hlitz: why is the code removed here? I thought we agree on setting the dcache cycle before op->state is set to OS_SCHEDULED

src/dcache_stage.c
```diff
@@ -287,10 +296,16 @@ void update_dcache_stage(Stage_Data* src_sd) {
       }
 
       op_set_done_cycle(op, cycle_count + DCACHE_CYCLES + op->inst_info->extra_ld_latency);
-      if (op->inst_info->table_info.mem_type != MEM_ST) {
+      // A value-predicted / RFP load already produced its register and woke its
+      // consumers at rename; do not wake (and re-produce) again here.
+      if (op->inst_info->table_info.mem_type != MEM_ST && !op->load_value_predicted) {
         op_set_wake_cycle(op, op_get_done_cycle(op));
         wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
       }
```
hlitz, line 307: assert in the else case that wake cycle was already set

src/dcache_stage.c
```diff
@@ -588,11 +603,15 @@ static inline void dcache_cacheline_hit(Op* op, Addr line_addr, Dcache_Data* lin
...
   }
 
-  /* wake up source inst if the op is completed */
-  if (op->inst_info->table_info.mem_type != MEM_ST) {
+  /* wake up source inst if the op is completed (skip if already produced early
+   * by value/RFP prediction at rename) */
+  if (op->inst_info->table_info.mem_type != MEM_ST && !op->load_value_predicted) {
     op_set_wake_cycle(op, op_get_done_cycle(op));
     wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
   }
```
hlitz, line 613: assert in the else case that wake cycle was set

src/dcache_stage.c
```diff
@@ -618,8 +637,15 @@ static inline void dcache_cacheline_miss(Op* op, Addr line_addr) {
         }
 
         op_set_done_cycle(op, cycle_count + DCACHE_CYCLES + op->inst_info->extra_ld_latency);
-        op_set_wake_cycle(op, cycle_count + DCACHE_CYCLES + op->inst_info->extra_ld_latency);
-        wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
+        // skip re-wake if produced early by value/RFP prediction at rename
+        if (!op->load_value_predicted) {
+          op_set_wake_cycle(op, cycle_count + DCACHE_CYCLES + op->inst_info->extra_ld_latency);
+          wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
+        }
```
hlitz, line 646: assert in else wake set

src/dcache_stage.c
```diff
@@ -879,10 +905,14 @@ static inline void dcache_fill_process_cacheline(Mem_Req* req, Dcache_Data* data
     }
     op->state = OS_SCHEDULED;
 
-    if (op->inst_info->table_info.mem_type != MEM_ST) {
+    // skip re-wake if produced early by value/RFP prediction at rename
+    if (op->inst_info->table_info.mem_type != MEM_ST && !op->load_value_predicted) {
       op_set_wake_cycle(op, op_get_done_cycle(op));
       wake_up_ops(op, REG_DATA_DEP, model->wake_hook);
     }
```
hlitz, line 913: assert wake cycle in else

src/decoupled_frontend.cc
```diff
@@ -1069,3 +1073,58 @@ void Decoupled_FE::redirect_to_off_path(FT_PredictResult result) {
   }
   ASSERT(proc_id, current_ft_to_push->get_end_reason() != FT_NOT_ENDED);
 }
+
+void Decoupled_FE::redirect_load_mispred(FT_PredictResult result) {
```
hlitz: lots of redundant code with redirect_to_off path. can we merge/refactor?

src/exec_stage.c
```diff
@@ -559,7 +559,7 @@ static inline void exec_stage_bp_resolve(Op* op) {
     bp_resolve_op(g_bp_data, op);
   }
 
-  if (op->bp_pred_info->recover_at_exec) {
+  if (op->bp_pred_info->recover_at_exec && !op->bp_pred_info->recovery_sch) {
```
hlitz: why is this change needed?

src/load_value_pred.cc
```diff
@@ -0,0 +1,606 @@
...
+ * resteering to the load's fall-through (op->oracle_info.npc) once the
+ * misprediction is detected; the recovery is *scheduled* later from the dcache
+ * stage (predicted_load_schedule_recovery), when the load reaches the point at
+ * which the prediction can be checked:
+ *   - recover_at_agen           : address prediction, checked at AGEN;
+ *   - recover_at_load_completion : value prediction, checked when data returns.
+ * On-path non-CF loads select bp_pred_main during predict_ft, so the flag lives
+ * there.  The reason is set at predict time; only one is ever set per load.
+ */
+static inline void load_pred_mark_recovery(Op* op, Flag recover_at_completion) {
```
hlitz: func can be inined/eliminated when simplified

src/exec_stage.c
```diff
@@ -329,9 +329,10 @@ void update_exec_stage(Stage_Data* src_sd) {
     /* update op status and execution metadata; increment PMU counters */
     exec_stage_process_op(op);
 
-    /* branch recovery/resolution */
+    /* branch recovery/resolution -- also fires for a mispredicted predicted load,
+     * which op_set_exec_cycle() marked recover_at_exec (recovers like a branch). */
     Flag is_replay = FALSE;  // TODO: check if this val is needed
-    if (op->inst_info->table_info.cf_type && !is_replay) {
+    if ((op->inst_info->table_info.cf_type || op->bp_pred_info->recover_at_exec) && !is_replay) {
```
hlitz: we said only to recover LVP at set_exec_cycle

src/exec_stage.c
```diff
@@ -551,7 +552,10 @@ static inline void exec_stage_process_op(Op* op) {
 }
 
 static inline void exec_stage_bp_resolve(Op* op) {
-  if (!BP_UPDATE_AT_RETIRE) {
+  // Only branches update branch-predictor state / take the branch-resolve latency
```
hlitz: undo all these changes and assert that bp_resolve is only called by CF ops, LVP are recovered elsehwere

src/op.h
```diff
@@ -314,6 +320,11 @@ static inline Counter op_get_exec_cycle(const Op* op) {
 static inline void op_set_exec_cycle(Op* op, Counter cycle) {
   ASSERT(op->proc_id, op->cycles.exec_cycle == MAX_CTR);
   op->cycles.exec_cycle = cycle;
+  // A mispredicted predicted load recovers like a branch, at exec: mark it here so
+  // exec_stage_bp_resolve schedules its squash (its recovery_info was filled at
+  // predict time). bp_pred_main is the level selected for on-path non-CF loads.
+  if (op->load_value_mispredicted && !op->off_path)
```
hlitz: bullshit. we only mark recover at exec in the decoupled frontend.

Here we check if recover_at_exec && ! CF_OP -> schedule_recovery. (we do not need use the bp_recovery which does a  lot of other stuff)

src/ft.cc
```diff
@@ -397,6 +398,30 @@ FT_PredictResult FT::predict_ft() {
...
+    // the load then recovers at exec like a branch (op_set_exec_cycle sets
+    // recover_at_exec, exec_stage_bp_resolve fires it). Here we only trigger the
+    // in-place off-path redirect after the load.
+    if (bp_id == MAIN_BP && op->inst_info->table_info.mem_type == MEM_LD) {
+      load_pred_predict_op(op);
+      if (op->load_value_mispredicted && !op->off_path) {
+        // The recovery resteers to the load's fall-through and squashes ops
+        // younger than the load; require the load to be its own end-of-macro op so
+        // no sibling uop is dropped. Value/address prediction targets such loads.
+        ASSERT(proc_id, op->eom);
```
hlitz: remove the assert on eom, recovery needs to be possible to be triggered by any op (not only eom).

Instead, in the recovery logic (triggered in set_exec_cycle) when recovery_at_exec executes schedule recover on the eom, by scanning op->parent_FT

src/dcache_stage.c
```diff
@@ -254,13 +254,11 @@ void update_dcache_stage(Stage_Data* src_sd) {
...
     // Record the first dcache access, keeping dcache_cycle write-once, before the
-    // state is set OS_SCHEDULED below (so op->state here is still the entry state).
-    // The only legitimate re-probe is a miss still waiting for a mem-request buffer
-    // (OS_WAIT_MEM), which stays resident and re-runs this access every cycle.
+    // state is set OS_SCHEDULED below. An op may re-probe this access across cycles
+    // (a miss waiting for a mem-request buffer in OS_WAIT_MEM re-runs it every cycle;
+    // predicted loads can also re-access), so only the first probe is recorded.
     if (op_get_dcache_cycle(op) == MAX_CTR)
       op_set_dcache_cycle(op, cycle_count);
-    else
```
hlitz, line 258: this is a crap fix. I don't understand the comment or what the exact ways are we can enter this code. 

but its clear that just checking against MAX_CTR is not good. I could set dcache cycle anywhere in the code at a random time and this would not catch it but just skip it.

src/ft.cc
```diff
@@ -400,24 +400,30 @@ FT_PredictResult FT::predict_ft() {
...
+        // than it, so scheduling on the EOM keeps the load's own sibling uops alive.
+        load_pred_mark_recovery(op);
+        // Go off-path after the macro, IN PLACE (no FT split): the consumers after
+        // the EOM are moved aside as the recovery FT and the FT's tail is refilled
         // with off-path ops. Driven from update() via redirect_load_mispred.
+        size_t eom_idx = idx;
+        while (eom_idx < ops.size() && !ops[eom_idx]->eom)
+          eom_idx++;
+        ASSERT(proc_id, eom_idx < ops.size() && ops[eom_idx]->inst_uid == op->inst_uid);
+        Op* eom = ops[eom_idx];
```
hlitz: why are we not using get_sibling_eom() here?

src/icache_stage.c
```diff
@@ -985,12 +985,13 @@ static inline void icache_process_ops(Stage_Data* cur_data, Flag fetched_from_uo
...
-    // (recover_at_*), so the following tail ops satisfy the
-    // ic->off_path == op->off_path invariant.
-    if (op->load_value_mispredicted && !op->off_path) {
+    // Load value/address mispredict: the decoupled frontend marked recover_at_exec
+    // on the mispredict's trigger uop and refilled this FT's tail off-path in place
+    // after the macro EOM. Flip the icache off-path right after the recovering
+    // macro's EOM (ft_get_sibling_eom(op) == op) -- not at the trigger, which may be
+    // mid-macro -- so the following tail ops satisfy the ic->off_path == op->off_path
+    // invariant, mirroring the branch flip above.
+    if (op->eom && op->inst_info->table_info.cf_type == NOT_CF && !op->off_path && ft_get_sibling_eom(op) == op) {
```
hlitz: isnt op->eom == (ft_get_sibling_eop(op) == op) ?

src/load_value_pred.cc
hlitz: why is recovery_scheduled needed?

src/map_rename.c
```diff
@@ -1027,9 +1028,13 @@ void reg_renaming_scheme_realistic_rename(Op *op) {
...
-  // checkpoint the speculative register table for recovering
-  if (!op->off_path &&
-      ((op->inst_info->table_info.cf_type && op->bp_pred_info->recover_at_exec) || op->load_value_mispredicted))
+  // checkpoint the speculative register table for recovering. A branch snapshots at
+  // its own rename (recover_at_exec). A predicted-load recovery snapshots at the
+  // macro EOM's rename (ft_get_sibling_eom(op) == op) so the checkpoint pairs with
+  // the EOM rollback, even though the trigger uop carrying recover_at_exec may be
+  // mid-macro.
+  if (!op->off_path && ((op->inst_info->table_info.cf_type && op->bp_pred_info->recover_at_exec) ||
+                        (op->eom && op->inst_info->table_info.cf_type == NOT_CF && ft_get_sibling_eom(op) == op)))
```
hlitz: eom == sibling?

src/decoupled_frontend.cc
```diff
@@ -1084,30 +1084,54 @@ void Decoupled_FE::redirect_load_mispred(FT_PredictResult result) {
...
+  // op->off_path invariant across the on->off boundary.
   ASSERT(proc_id, ops[eom_idx] == result.op);
-  saved->op_pos = 0;
-  saved->generate_ft_info();
-  saved->set_prebuilt(true);
-  saved_recovery_ft = saved;
 
-  // Refill the SAME FT's tail with off-path ops (no split), then continue
-  // off-path -- identical to redirect_to_off_path's off-path fill, but in place.
+  if (eom_idx + 1 < ops.size()) {
```
hlitz: the following code has lots of redundancy with bp mispred handling. can we refactor out common code or so?

src/ft.cc
```diff
@@ -397,6 +398,38 @@ FT_PredictResult FT::predict_ft() {
...
+        load_pred_mark_recovery(op);
+        // Go off-path after the macro, IN PLACE (no FT split): the consumers after
+        // the EOM are moved aside as the recovery FT and the FT's tail is refilled
+        // with off-path ops. Driven from update() via redirect_load_mispred. Resolve
+        // the recovery point (macro EOM) through the same ft_get_sibling_eom used by
+        // the scheduler/snapshot, then locate its index for the in-place split.
+        Op* eom = ft_get_sibling_eom(op);
+        ASSERT(proc_id, eom);
+        size_t eom_idx = idx;
+        while (ops[eom_idx] != eom)
```
hlitz: what is this doing? we are setting it to eom then increment and check if we find another eom. ??

src/ft.cc
```diff
@@ -535,6 +568,37 @@ FT_Info ft_get_ft_info(FT* ft) {
...
+  bool recovers = false;
+  for (Op* cand : ft->get_ops()) {
+    if (!cand || cand->inst_uid != op->inst_uid)
+      continue;
+    if (cand->eom)
+      eom = cand;
+    if (cand->bp_pred_main.recover_at_exec)
+      recovers = true;
+  }
+  return recovers ? eom : NULL;
```
hlitz: why do we check for recovers? isn't it sufficient to ahve an API that always returns the eom of a given op. we want this API both to schedule recovery on the eom and we need it to determine when to do SRT but this should be independent of whether it recovers.

src/decoupled_frontend.cc
```diff
@@ -1047,25 +1077,61 @@ void Decoupled_FE::redirect_to_off_path(FT_PredictResult result) {
...
+  FT* ft = current_ft_to_push;
+  const uint64_t eom_idx = result.index;
+  std::vector<Op*>& ops = ft->get_ops();
+  ASSERT(proc_id, eom_idx < ops.size());
+  // The macro EOM is the recovery point. The icache flips off-path right after
+  // fetching it (see ft_get_sibling_eom), keeping the per-op ic->off_path ==
+  // op->off_path invariant across the on->off boundary.
+  ASSERT(proc_id, ops[eom_idx] == result.op);
+
+  if (eom_idx + 1 < ops.size()) {
```
hlitz: why not check eom_idx == ops.size() ?

src/decoupled_frontend.cc
```diff
@@ -1047,25 +1077,61 @@ void Decoupled_FE::redirect_to_off_path(FT_PredictResult result) {
...
+    std::vector<Op*>& saved_ops = saved->get_ops();
+    for (size_t i = eom_idx + 1; i < ops.size(); i++) {
+      ops[i]->parent_FT = saved;
+      saved_ops.push_back(ops[i]);
+    }
+    ops.resize(eom_idx + 1);
+    saved->op_pos = 0;
+    saved->generate_ft_info();
+    saved->set_prebuilt(true);
+    saved_recovery_ft = saved;
```
hlitz: can we do it more similar as in the bp case? i think there we just copy the current ft into saved and we build a new one if ncesseray

src/ft.cc
```diff
@@ -568,35 +566,41 @@ FT_Info ft_get_ft_info(FT* ft) {
...
- *     trigger is mid-macro. */
+/* Return the END-OF-MACRO op of op's macro-instruction (the uop with the same
+ * inst_uid and eom set), by scanning op's parent FT. Pure: always returns the EOM,
+ * independent of whether the macro recovers. A predicted load's recovery is
+ * TRIGGERED by a (possibly mid-macro) uop's execution but must be SCHEDULED on the
+ * EOM (bp_sched_recovery squashes ops younger than the recovery op, so the EOM keeps
+ * the load's own sibling uops alive). The rename SRT snapshot and icache off-path
+ * flip use "op->eom" for the same "op is the EOM" test (equivalent for a valid
+ * macro) and gate recovery separately via ft_sibling_recovers_at_exec. */
 Op* ft_get_sibling_eom(Op* op) {
```
hlitz: have this api return both the eom op and the eom_op_idx so we don not have to do it again

src/icache_stage.c
```diff
@@ -986,13 +986,12 @@ static inline void icache_process_ops(Stage_Data* cur_data, Flag fetched_from_uo
...
-    // of a recovering macro" (it returns NULL unless some uop of the macro carries
-    // recover_at_exec), so flip the icache off-path right after that EOM -- even when
+    // on the mispredict's trigger uop and refilled this FT's tail off-path after the
+    // macro EOM. Flip the icache off-path right after that EOM (op->eom) -- even when
     // the trigger is mid-macro -- keeping the ic->off_path == op->off_path invariant.
-    // cf_type == NOT_CF excludes branch-recovery macros, which flip above.
-    if (!op->off_path && op->inst_info->table_info.cf_type == NOT_CF && ft_get_sibling_eom(op) == op) {
+    // ft_sibling_recovers_at_exec gates it to recovering macros; cf_type == NOT_CF
+    // excludes branch-recovery macros, which flip above.
+    if (!op->off_path && op->eom && op->inst_info->table_info.cf_type == NOT_CF && ft_sibling_recovers_at_exec(op)) {
```
hlitz: why do we need this? Why do we not just immetdiately go on the wrong path as soon as we detect a recover_at_exec (or maybe even op->off_path is enough) why do you want to flip it at eom?

src/map_rename.c
```diff
@@ -1030,11 +1030,11 @@ void reg_renaming_scheme_realistic_rename(Op *op) {
...
   // its own rename (recover_at_exec). A predicted-load recovery snapshots at the
-  // macro EOM's rename (ft_get_sibling_eom(op) == op) so the checkpoint pairs with
-  // the EOM rollback, even though the trigger uop carrying recover_at_exec may be
-  // mid-macro.
+  // macro EOM's rename (op->eom) -- the recovery point -- so the checkpoint pairs
+  // with the EOM rollback even though the trigger uop carrying recover_at_exec may
+  // be mid-macro; ft_sibling_recovers_at_exec gates it to recovering macros.
   if (!op->off_path && ((op->inst_info->table_info.cf_type && op->bp_pred_info->recover_at_exec) ||
-                        (op->inst_info->table_info.cf_type == NOT_CF && ft_get_sibling_eom(op) == op)))
+                        (op->eom && op->inst_info->table_info.cf_type == NOT_CF && ft_sibling_recovers_at_exec(op))))
```
hlitz: instead of sibling can we introduce map_offpath which is set 

if (op->off_path && op->eom && !map_offpath) {
map_offpath = true;
reg_file_snaphot_srt();
}

and in recover we need to set map_offpath = false;

src/icache_stage.c
hlitz: if we really need it just check for if(op->off_path && !ic->off_path && op->eom)

hlitz: get rid of the sibling stuff

src/decoupled_frontend.cc
```diff
@@ -1047,22 +1077,39 @@ void Decoupled_FE::redirect_to_off_path(FT_PredictResult result) {
...
-    }
+  refill_ft_tail_off_path(current_ft_to_push);
+  if (current_ft_to_push->ended_by_exit()) {
+    next_state = INACTIVE;
+    exit_on_off_path = true;
   }
+  ASSERT(proc_id, current_ft_to_push->get_end_reason() != FT_NOT_ENDED);
+}
+
+void Decoupled_FE::redirect_load_mispred(FT_PredictResult result) {
```
hlitz: i still think this is garbage. what is the exact difference between redirect_load_mispred and redirect_to_offpath ?

if the change is very small just add a parameter to redirect_to_offpath(bool caused_by_lvp) and then handle the diffference within a single func

PR #554, opened 2026-07-27
Poison oracle memory addresses in the trace replay buffer

hlitz:
## Problem

The trace frontend keeps a PC-keyed map (`pc_to_inst`) of previously seen on-path instructions and replays them to synthesize wrong-path instructions in `off_path_generate_inst()` (`src/frontend/pt_memtrace/trace_fe.cc`). Each stored entry carries the `ld_vaddr`/`st_vaddr` recorded at the **on-path** occurrence of that PC.

Replaying those addresses off-path leaks oracle information: wrong-path loads and stores touch exactly the lines the correct path will later need, so speculative memory behavior acts as a perfectly accurate prefetcher rather than as the mostly-wrong accesses real hardware would issue.

## Change

Overwrite the oracle addresses of every replay-buffer entry with a non-canonical poison value (`0xCAFEBABEDEADBEAF`), gated by a new parameter:

```
DEF_PARAM(poison_replay_buffer_oracle_va, POISON_REPLAY_BUFFER_ORACLE_VA, Flag, Flag, TRUE, )
```

Default is **on**, so wrong-path memory behavior is not oracle-informed unless explicitly re-enabled with `--poison_replay_buffer_oracle_va 0`.

## Measured impact

SPEC CPU2017 memtrace simpoints, `sunny_cove`, `--lookahead_buf_size 256`, 149 simpoints compared three ways against an unmodified baseline:

| config | mean Δ cycles vs baseline |
| --- | --- |
| `--lookahead_buf_size 256` | **−0.356%** |
| same + poisoned oracle VAs | **+0.060%** |

On the 25 simpoints where the lookahead buffer won most (>0.5% fewer cycles), the mean goes from **−1.944%** to **+0.310%**. Representative cases:

```
blender_r/95521    -4.26%  ->  +1.88%
gcc_r_3/1264       -4.83%  ->  +0.24%
nab_r/59950        -1.99%  ->  +0.03%
wrf_r/243792       -1.43%  ->  +0.03%
```

The lookahead buffer's apparent speedup was therefore largely an oracle artifact of replayed wrong-path addresses, not a fetch-side effect.

## Known issue

With poisoning enabled, the `ctype_pin_inst_same_mem_vaddr()` check in `ext_trace_fetch_op()` compares real incoming addresses against stored poisoned ones, so it always mismatches for memory instructions. This causes a redundant erase+reinsert and an `INST_MAP_UPDATE_MEM_*` event per repeat memory op. Functionally harmless — the entry is rewritten with poison either way — but it makes those stats meaningless while the flag is on. Left as-is to keep the diff minimal; happy to fix here or in a follow-up.

### DIFF
--- a/src/core.param.def
+++ b/src/core.param.def
@@ -335,6 +335,15 @@ DEF_PARAM(l1_miss_rate, L1_MISS_RATE, uns, uns, 10, )
 
 DEF_PARAM(trace_buf_size, TRACE_BUF_SIZE, uns, uns, 0, )
 
+/* Replace the oracle memory addresses (ld_vaddr/st_vaddr) of every instruction
+   recorded in the trace frontend's replay buffer with a poison value. The
+   replay buffer is keyed by PC and is replayed to synthesize wrong-path
+   instructions; the addresses it stores come from the on-path occurrence of
+   that PC, so reusing them off-path leaks oracle information and makes
+   wrong-path memory accesses behave like a perfect prefetcher. Enabled by
+   default so wrong-path memory behavior is not oracle-informed. */
+DEF_PARAM(poison_replay_buffer_oracle_va, POISON_REPLAY_BUFFER_ORACLE_VA, Flag, Flag, TRUE, )
+
 DEF_PARAM(perfect_confidence, PERFECT_CONFIDENCE, Flag, Flag, FALSE, )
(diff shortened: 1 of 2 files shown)


yql5510718 (approved):
LGTM.
Thank you for tracking this down. This has been haunting us for some time.
Still see some noise level IPC change. I guess that is from actual off-path instruction instead of nop.

PR #553, opened 2026-07-27
refactor: Op cycle counters into a struct with get/set accessors

hlitz:
## Summary

Refactors the per-op pipeline **cycle counters** into a single struct accessed
through a typed get/set API, so per-counter behavior and invariants can live in
one place. Based directly on `main` (independent of the LVP work).

## What changed

1. **`Op_Cycles` struct** (op.h) holds all 15 counters — `fetch, bp, map, issue,
   rdy, sched, exec, dcache, done, retire, replay, pred, precommit, decode, wake`.
   The flat fields (including the stray `wake_cycle` that lived in the dependency
   block) are removed from `Op` and replaced by `Op_Cycles cycles`.

2. **Per-counter accessors** — explicit `op_get_<name>_cycle()` /
   `op_set_<name>_cycle()` functions (one each, not a macro, so each can grow
   custom logic). Every setter is **write-once**: it asserts the counter is still
   `MAX_CTR` (its post-allocation sentinel). Multiple assignment *sites* are fine
   as long as they are mutually exclusive for a given op.

3. **`rdy_cycle` exception** — it is a MAX-over-producers accumulator (written
   once per waking source), so `op_set_rdy_cycle` has no assert and it resets to
   `1` (the max identity / born-ready default), not `MAX_CTR`.

4. **`op_assert_cycles_set_at_retire()`** — at retirement every on-path op must
   have all counters set. `bp_cycle` is checked only for CF ops, `dcache_cycle`
   only for memory ops; `replay_cycle`, `pred_cycle`, and `retire_cycle` are
   excluded (conditional / set at retirement).

5. All reads/writes converted tree-wide. `rdy_cycle`/`fetch_cycle` collide with
   non-Op structs (`Mem_Req`, `Dcache_Data`, `Icache_Data`, `Op_Info`,
   prefetchers); those (`req->`, `data->`, `line->`, `line_info->`) are left
   untouched. `op_pool.c` and `uop_generator.c` reset the struct directly (now
   including `decode_cycle`); `idq_stage.c`'s `!decode_cycle` test becomes
   `== MAX_CTR`. `param_parser.c` gets `#undef ASSERTM` before its soft
   redefinition (op.h now pulls `globals/assert.h`).

## Status
- Builds clean (opt-avx); clang-format applied to changed lines.

## Known / for discussion
The write-once setters for **`exec`/`done`/`wake`** will trip at runtime on
**replay / re-execution** — those counters are re-set on a second pipeline pass
[...]

PR #556, opened 2026-07-30
dcache: skip re-lookup for OS_WAIT_MEM re-probes + tighten dcache_cycle stamp

hlitz:
Follow-up to #553 (based on `refactor-cycle-counters`; rebase onto `main` once #553 lands). Single commit.

## What
A memory op that misses and can't get a mem-request buffer parks in `OS_WAIT_MEM` and re-probes the dcache every cycle. `new_mem_req` searches all queues and **merges before** it ever returns "no buffer", so `OS_WAIT_MEM` means *a known miss with no in-flight request to merge into*. The old re-probe still re-ran the full access — acquire a port, look up the tag, re-derive the same miss — burning a dcache port each cycle (and able to bounce the op through `OS_WAIT_DCACHE`) without ever changing the outcome.

This short-circuits `OS_WAIT_MEM` re-probes straight into the miss handler (retry `new_mem_req` only), skipping the port acquisition and the tag lookup. `line_addr` comes from the side-effect-free `get_cache_line_addr`. A merge partner that appears later is still caught by `new_mem_req`'s own queue search. `OS_SCHEDULED` is set first (as the fresh-access path does) so the store-forward-hit case — which leaves `op->state` untouched — stays removable rather than resident.

## Why it also tightens the dcache_cycle stamp
Removing the re-probe eliminates the **only** path by which a stamped op re-enters `OS_WAIT_DCACHE` (`OS_WAIT_MEM` → busy port → `OS_WAIT_DCACHE`). So every op reaching the stamp site is now on its unique first access, and the guard tightens from "stamp-if-unset, else assert a re-probe state" (the #553-safe form) back to strict "assert unset, set once."

## Not a no-op
Fewer dcache port acquisitions + no redundant re-lookup → this changes timing. Validated for IPC drift on the full SPEC2017 memtrace sweep (`json/spec17_skip_relookup.json`, baseline = the #553 binary) — results to follow in a comment. Kept separate from #553 precisely because #553 must stay a functional no-op.

### DIFF
--- a/src/dcache_stage.c
+++ b/src/dcache_stage.c
@@ -234,6 +234,21 @@ void update_dcache_stage(Stage_Data* src_sd) {
       continue;
     }
 
+    // OS_WAIT_MEM re-probe: the op already accessed the dcache and missed, and
+    // new_mem_req found no in-flight request to merge with and no free mem-request
+    // buffer. Re-looking-up the dcache can only re-derive the same known miss, so skip
+    // the port acquisition and the tag lookup entirely and go straight to retrying the
+    // mem request. A matching in-flight request that shows up later is still caught by
+    // new_mem_req's queue search inside the miss handler. Set OS_SCHEDULED first (as the
+    // fresh-access path below does) so the store-forward-hit case, which does not set
+    // op->state, leaves the op removable instead of resident.
+    if (op->state == OS_WAIT_MEM) {
+      ASSERT(dc->proc_id, op_get_dcache_cycle(op) != MAX_CTR);
+      op->state = OS_SCHEDULED;
+      dcache_cacheline_miss(op, get_cache_line_addr(&dc->dcache, op->oracle_info.va));
+      continue;
+    }
+
     /* check on the availability of a read port for the given bank */
     // the bank bits are the lowest order cache index bits
     uns bank = BANK(op->oracle_info.va, DCACHE_BANKS, DCACHE_INTERLEAVE_FACTOR);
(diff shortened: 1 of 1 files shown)


src/dcache_stage.c
```diff
@@ -245,18 +260,14 @@ void update_dcache_stage(Stage_Data* src_sd) {
...
-      op_set_dcache_cycle(op, cycle_count);
-    else
-      ASSERT(dc->proc_id, op->state == OS_WAIT_DCACHE || op->state == OS_WAIT_MEM);
+    // Record the op's first dcache access (dcache_cycle gates precommit). Every op
+    // reaching here is making that first access: fresh from the scheduler, or an
+    // OS_WAIT_DCACHE op that was only ever waiting for a cache port and never accessed.
+    // OS_WAIT_MEM re-probes are short-circuited above, so a stamped op can no longer
+    // re-enter this point (the sole path was OS_WAIT_MEM -> busy port -> OS_WAIT_DCACHE).
+    // dcache_cycle must therefore be unset.
+    ASSERT(dc->proc_id, op_get_dcache_cycle(op) == MAX_CTR);
```
moonseoppkim: I think this is redundant. we are checking this inside the "op_set_dcache_cycle" function.

hlitz: Good catch — removed the redundant ASSERT; `op_set_dcache_cycle` already enforces write-once. Fixed in the amended commit.

moonseoppkim (approved):
LGTM

PR #561, opened 2026-08-09
[BUGFIX] ISSUE_WIDTH param adjust

yzhao321:
Scarab defines ISSUE_WIDTH as the width of renaming and dispatching, which is incorrect.
This patch aims at fixing this definition, by adding the params of DISPATH_WIDTH and RENAME_WIDTH. ISSUE_WIDTH can be used by NUM_FUS.


hlitz (approved):
LGTM

PR #562, opened 2026-08-11
2.4–4.1% Runtime reduction using Intrusive List

Kaden123Lee:
**Synopsis:** The main costs were the separate allocation and deallocation of std::list nodes and the O(n) search performed by std::list::remove(). The intrusive-list implementation embeds the links directly in IssueQueueEntry, eliminating the separate list-node allocation and making removal O(1).

1. entry points to an IssueQueueEntry.
2. The list stores that pointer as a value inside a separate list node.
3. remove(entry) searches O(n) for a list node whose stored pointer equals entry.
4. It rewrites that node’s neighbors.
5. It destroys and deallocates the list node.
6. The IssueQueueEntry remains alive.

New intrusive version:
1. entry still points to the same IssueQueueEntry.
2. That object contains its own ready_prev and ready_next.
3. Removal directly rewrites the neighboring entries’ pointers in O(1).
4. It clears its own links and marks itself absent.
5. No separate list node is destroyed or deallocated.
7. The existing IssueQueueEntry storage can be reused later.

Each configuration was run five times per workload.
--------------------------------------------------------------------------------------------------------
perlbench baseline:  172.809 176.254 179.626 179.016 174.971
perlbench intrusive: 171.727 172.098 167.382 174.223 173.108

mcf baseline:        189.910 189.889 188.556 188.726 186.323
mcf intrusive:       183.181 184.945 181.768 182.377 183.901

lbm baseline:        153.271 156.487 156.928 157.641 156.354
lbm intrusive:       148.014 150.310 151.904 151.468 149.615

- 570 of 570 simulator-output comparisons matched excluding SIM_HOST_WALL_SECONDS
- IPC is identical.
- Simulated instruction counts are identical.
- Simulated cycle counts are identical.
- Cache, branch predictor, memory, prefetcher, power, and Ramulator statistics are identical.
- All 15 PARAMS.out comparisons matched
-------------------------------------------------------------------------------------------------------
Notes:
image
image
image
image
image
image
image
image
image
image
image

**Future work:** I’ve made two optimizations to the structure of a data container, so now I’m thinking about making bigger changes that could improve speed. I plan to learn some new tools to help me find similar structures. For example, if I have 15 files that could all use small-size optimized vectors to boost speed or memory performance, I could create one shared structure and use it across those files. I think this is the next step in my learning.

I'll also be fixing my bug fix, but I wanted to work on some heaptrack stuff.

### DIFF
--- a/src/issue_queue.cc
+++ b/src/issue_queue.cc
@@ -53,7 +53,6 @@ extern "C" {
 
 #include <algorithm>
 #include <deque>
-#include <list>
 #include <memory>
 #include <sstream>
 #include <string>
@@ -114,6 +113,10 @@ struct IssueQueueEntry {
 
   uns32 bound_fu_id = MAX_UNS;
   ISSUE_QUEUE_ENTRY_STATE state = ISSUE_QUEUE_ENTRY_STATE_EMPTY;
+
+  IssueQueueEntry* ready_prev = nullptr;
+  IssueQueueEntry* ready_next = nullptr;
+  bool in_ready_list = false;
 
   explicit IssueQueueEntry(uns16 queue_id, uns16 entry_id) : queue_id(queue_id), entry_id(entry_id) {}
   void clear();
@@ -273,7 +276,7 @@ class SelectLogic {
   std::unique_ptr<BindPolicy> bind_policy;
 
   std::vector<size_t> picker_order;  // picker traversal order generated each cycle
-  std::list<IssueQueueEntry*> ready_list;
+  IssueQueueEntry* ready_head = nullptr;
 
   // bitmask of ready but not yet issued op types in this cycle
   uns64 ready_not_issued_op_types = 0;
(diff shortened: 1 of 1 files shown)


hlitz:
Hi @yzhao321 didn't we discuss that exact option before? I remember I was proposing before to eliminate the ready list and just scan the IQ array because scanning arrays is usually faster than pointer chasing. But you said iterating over the ready_list (because it is smaller) turned out to be faster. So I don't think we need the ready-list for number of ready ops only but we also need it for bid(). I think iterating over a ready bitvector would be kind of the same as iterating over the IQ itself.

In any case we need perf data. I would like to see again the perf of just dropping the ready_list and iterating over the IQ directly.

If the readylist bitvector is very sparse, it maybe worth to copy readylist, find-first-bit-set, then mask out that bit but not sure.

Kaden123Lee:
@hlitz I submitted the BSOE account request with you as my sponsor. It says you should have received an email asking for confirmation.

yzhao321:
> Hi @yzhao321 didn't we discuss that exact option before? I remember I was proposing before to eliminate the ready list and just scan the IQ array because scanning arrays is usually faster than pointer chasing. But you said iterating over the ready_list (because it is smaller) turned out to be faster. So I don't think we need the ready-list for number of ready ops only but we also need it for bid(). I think iterating over a ready bitvector would be kind of the same as iterating over the IQ itself.
> 
> In any case we need perf data. I would like to see again the perf of just dropping the ready_list and iterating over the IQ directly.
> 
> If the readylist bitvector is very sparse, it maybe worth to copy readylist, find-first-bit-set, then mask out that bit but not sure.

Hi @hlitz
We did discuss it before.
At first I proposed that using the order of the ready for bid. The expr shows that maintaining the order is too expensive. So we got rid of this scheme.
Then actually we don't need the ready list any more. I did three impl: 1. Scan the whole issue queue without ready list; 2. have a ready list using self-define struct; 3. have a ready list using std::list. At that time, there are not too much performance differences and the perf report did not show the hotspot.
Then we just picked scheme 3 for in the future we may use the ready list? We do not need ready list for bid as we can always simply scan the whole issue queue.

